PR #5, opened 2016-02-08
Merge osxfuse

jmmv:
@Nikratio @bfleischer: As we were discussing in https://github.com/osxfuse/sshfs/pull/20, it would be good to merge osxfuse's fork of sshfs into libfuse because the latter is considered to be the "main copy".

This pull request does that. I pulled osxfuse's sshfs branch into a local branch, ran a merge on top of libfuse's code and then added a bunch of commits of my own to clean up the results. You'll see that the changes are minimal and there should be no reason to keep them in a fork.

This should be a no-op for Linux because the majority of the code is protected behind `__APPLE__` preprocessor conditionals. I don't like this because checks should be done via configure, but I wanted to just merge the code with as few changes as possible to preserve current behavior. We can improve things separately.

The best way to review this PR is to click on *Files changed* and see the whole diff for the pre-merge and post-merge trees. The changes are minimal. Attempting to review all individual commits is difficult (as I encountered when trying to do this with a rebase instead of a merge).


Makefile.am
```diff
@@ -6,16 +6,24 @@ sshfs_SOURCES = sshfs.c cache.c cache.h
...
-CLEANFILES = sshnodelay.so
+EXTRA_DIST = sshnodelay.c FAQ.txt
+CLEANFILES = sshnodelay.so sshfs.1 sshfs.1.tmp
 
 dist_man_MANS = sshfs.1
 
+sshfs.1: sshfs.1.in
+	$(AM_V_GEN)$(CPP) $(CPPFLAGS) -P -xassembler-with-cpp sshfs.1.in \
+	    | sed -e '/^$$/d' >sshfs.1.tmp || exit 1; \
+	mv sshfs.1.tmp sshfs.1
```
Nikratio: Ugh. This is very hacky (treat groff text as assembler with CPP) and very confusing (`.in` is the extension used for autoconf templates). Can't we use autoconf, so that sshfs.1 is generated from sshfs.1.in at configure time?

Makefile.am
```diff
@@ -6,16 +6,24 @@ sshfs_SOURCES = sshfs.c cache.c cache.h
...
+sshfs_SOURCES += compat/darwin_compat.c compat/darwin_compat.h
+endif
 
 sshfs_LDADD = $(SSHFS_LIBS)
 sshfs_CFLAGS = $(SSHFS_CFLAGS)
 sshfs_CPPFLAGS = -D_REENTRANT -DFUSE_USE_VERSION=26 -DLIBDIR=\"$(libdir)\"
 
-EXTRA_DIST = sshnodelay.c
-CLEANFILES = sshnodelay.so
+EXTRA_DIST = sshnodelay.c FAQ.txt
```
Nikratio: Does this work? There's no FAQ file nor is there code to create it. At some point FAQ.txt was generated by converting on of the sshfs wiki pages, but that code has been removed.

compat/darwin_compat.c
```diff
@@ -0,0 +1,246 @@
...
+    pthread_mutex_unlock (&sem->__data.local.count_lock);
+
+    return res;
+}
+
+/* http://www.opengroup.org/onlinepubs/007908799/xsh/sem_wait.html */
+int
+darwin_sem_wait(darwin_sem_t *sem)
+{
+    int res = 0;
```
Nikratio: At least under Linux, this needs to be volatile: `pthread_cleanup_*` may call longjmp, which then clobbers the value of `res`. Unless you're 100% sure that this isn't a problem with Darwin's pthread implementation, I'd recommend making this `volatile` as well.

sshfs.c
```diff
@@ -190,6 +204,9 @@ struct sshfs_file {
 	int connver;
 	int modifver;
 	int refs;
+#ifdef __APPLE__
+	pthread_mutex_t file_lock;
```
Nikratio: Why does this need to be locked under Darwin, but not elsewhere?

sshfs.c
```diff
@@ -849,11 +879,35 @@ static void ssh_add_arg(const char *arg)
...
+	snprintf(sopath, sizeof(sopath), "%s/%s", sshfs_program_path_base,
+             SSHNODELAY_SO);
+	res = access(sopath, R_OK);
+	if (res == -1) {
+		goto nobundle;
+	}
+	goto pathok;
+
+nobundle:
+#endif /* __APPLE__ */
```
Nikratio: No idea what's happening here - I guess this is OS-X specific?

sshfs.c
```diff
@@ -3494,6 +3596,15 @@ static int read_password(void)
 		perror("Failed to allocate locked page for password");
 		return -1;
 	}
+#ifdef __APPLE__
```
Nikratio: Why only under OS-X? Looks like a good idea in general.

sshnodelay.c
```diff
@@ -5,6 +5,32 @@
...
+
+typedef struct interpose_s {
+	void *new_func;
+	void *orig_func;
+} interpose_t;
+
+static const interpose_t interposers[] \
+	__attribute__ ((section("__DATA, __interpose"))) = {
+	{ (void *)custom_connect,  (void *)connect  },
+};
```
Nikratio: Any chance to get a brief comment what's happening here non OS-X people?

Nikratio:
Other than the above, this looks fine to me. Thanks for working on this!

Makefile.am
jmmv: No, it didn't. I have removed this and ensured that `make distcheck` works (which required other fixes and seems to have been already broken in HEAD).

Makefile.am
jmmv: Indeed it's an ugly hack. I didn't want to touch the code much as part of the merge... but considering this is brand new, I have redone this.

I've still kept the generation logic in the `Makefile` though. I personally think that using `AC_OUTPUT` to generate arbitrary files that are not part of the build system is wrong. Artifacts of the build should be generated by `Makefile` rules to avoid having to go through the possibly-expensive `configure` cycle just to regenerate such outputs. `Makefile`s are better at tracking finer-grained dependencies.

compat/darwin_compat.c
jmmv: I don't know. Do you recommend doing this for this single instance of res, or for others as well? I'd imagine `trywait`, `timedwait` and possibly `getvalue` are also affected.

sshfs.c
jmmv: I don't know and this may be hard to track down.

I see this comes from b4023a19dd7ec7a099d2e0df491547cf3bb6bec3, which is a very old commit that says nothing more than `Apply modified sshfs 2.2 patch from MacFUSE`. I guess the locking semantics in MacFUSE might have been different at the time (though that'd have been stupid for compatibility).

I've removed this for now assuming it's a hack that's no longer necessary. I have ran a few local tests and things seem stable. Those tests are not very scientific... but sshfs has no test suite anyway. I can try a much stronger workload tomorrow at work to better assess this.

sshfs.c
jmmv: Agreed, so done. I guess this is only because MacFUSE wanted to protect everything under `__APPLE__` to keep their changes in the fork isolated and easier to look for.

sshfs.c
jmmv: I don't know for sure, but "bundles" in OS X are a way of shipping multiple files as a single entity. The UI displays such entities as regular files (and they are handled as a unit), but underneath they are actually directories with a specific file layout.

I suppose that either MacFUSE or OSXFUSE are distributing sshfs as a bundle and that within the bundle is the `sshnodelay.so` file. This trick allows for the library to be found within the bundle instead of in the system path.

sshnodelay.c
jmmv: Sure. I was going to say "don't know", but then I had to look around and found: http://toves.freeshell.org/interpose/interpose_old.html . Pretty interesting.

I have added a comment and taken the chance to remove code duplication in this file.

(Haven't build-tested this specific change on Linux yet; will have to do so tomorrow from work.)

jmmv:
OK, I think I have addressed all your comments. I'll test the stability of this tomorrow at work where I have a pretty intense scenario to run on sshfs. PTAL

compat/darwin_compat.c
Nikratio: `trywait` and `getvalue` don't use `pthread_cleanup_*`, so they are good. `timedwait` is an interesting case, looking at the code it should need the same fix. But for `sem_wait` I actually got a compiler warning as well - for `timedwait` I don't. I think I'd add it to be on the safe sid.

jmmv: `pthread.h` in OS X implements cleanup handlers via macros and they handle state via the `pthread_self()` opaque object. I have found zero mentions of `setjmp` et. al. on the pthread headers nor in the source in http://www.opensource.apple.com/source/Libc/Libc-391/pthreads/ . So I think the code is safe as is.

Now, there is the question on why this reimplementation of semaphores exists at all: current versions of OS X do have the `sem_*` primitives we need in sshfs. What's more is that they are deprecated, so presumably they've been around for a long time.

I'd rather not touch this at this point. My take is that merging the codebases "as they are" is better than attempting to refactor stuff at the same time: doing a merge first will allow for osxfuse's sshfs fork to wind down easily. If we change too many things as part of the merge and something goes wrong in the next release of sshfs, it'd be too easy to say that libfuse-sshfs is unstable on OS X and that people should continue using osxfuse-sshfs instead.

jmmv:
Alright. I have done some intensive testing with a sshfs built from this branch and things look stable. (If you are curious, I built a rather large source tree of over 1GB mounted over sshfs.)

compat/darwin_compat.c
Nikratio: OS-X does not yet have a full semaphore implementation, e.g. `sem_init()` is missing.

jmmv: Are you sure? Removing the compat file and just relying on OS X's `semaphore.h` allowed this to build.

jmmv:
Thanks!

compat/darwin_compat.c
Nikratio: I'm sure that OS-X doesn't have `sem_init` (I actually re-used sshfs' compatibility layer in another project for exactly that reason). I didn't say that sshfs itself uses `sem_init` :-).

If it builds (and works) fine on your system, then we should get rid of the layer completely.

jmmv:
```
$ nm /usr/lib/system/libsystem_kernel.dylib | grep sem_
000000000001719c T ___sem_open
00000000000171b4 T ___sem_wait_nocancel
000000000001852c T _sem_close
00000000000140db T _sem_destroy
00000000000140f1 T _sem_getvalue
0000000000014107 T _sem_init
000000000001719c T _sem_open
0000000000018544 T _sem_post
000000000001855c T _sem_trywait
0000000000018574 T _sem_unlink
000000000001858c T _sem_wait
00000000000171b4 T _sem_wait$NOCANCEL
```

So... it does build but it does not work later, and I haven't investigated. Note also that the `sem_*` functions in OS X are all marked as deprecated so maybe this is not a good route to take.

I've also noticed that osxfuse provides the `sem_*` compatibility layer on its own, so we could probably use that instead of using this local copy. Need to look into this a bit more to see which minimum version of osxfuse we'd need because what I see installed on my system does not match the contents of the osxfuse repository...

bfleischer: The `sem_*` compatibility function are needed. The compatibility layer offered by osxfuse is deprecated and will be gone in future releases. osxfuse 3 still contains the function, but not the corresponding headers. In osxfuse 4 the function will be gone, too.

bfleischer: @jmmv The osxfuse repository contains a branch for osxfuse 2.x and a separate branch for the upcoming osxfuse 3.x. You might be looking at the wrong branch.

jmmv: Thanks for the explanation, @bfleischer . Have you done any experimentation using the native `sem_*` calls though?

Nikratio: Trying to call `sem_init` fails with errno 78, "Function not implemented".

PR #19, opened 2016-06-02
Implement support for "fsync@openssh.com"

rianhunter:
OpenSSH implements fsync() via the extension "fsync@openssh.com". This
change uses that extension when receiving a FUSE fsync request.


rianhunter:
The second commit is unrelated but you may pull that in as well :)

PR #20, opened 2016-06-06
Require paths for truncate/fstat workarounds

rianhunter:
Commit 74bfa3850a2568f96dd1d090a9386534c9bb4629 allowed sshfs to run
without requiring a non-NULL path argument for certain
operations. This was erroneous in the case of using certain
workarounds.


rianhunter:
Ping

Nikratio:
sorry, thanks for the patch!

PR #22, opened 2016-06-23
add parentheses to suppress warning

iori-yja:
With gcc 6.1.1, warning message emerges.

```
sshfs.c: In function ‘sshfs_fsync’:
sshfs.c:2643:2: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
  if (err = sshfs_flush(path, fi))
  ^~
```

`if (assignment) pattern' sometime happens mistakenly, and the best way to tell the compiler that the sentence is not made by mistake is adding parentheses around assignment.


Nikratio:
Please also add a comment explaining the double parentheses. Otherwise this may make the compiler happy, but it would make me suspicious when looking at the code.

iori-yja:
Sorry for late, it's been done

Nikratio:
Sorry for the delay.. would you have time to rebase against current master? The patch doesn't apply cleanly anymore.

iori-yja:
I've found the @Nikratio's comment as I zapped my page that happens hardly ever, so it is so lucky to find a comment only 4 days dated.., anyway though.

I've merged recent upstream, so I think you would be easy to merge, wouldn't you? Sorry for being rude that pull-requesting to master.

Nikratio:
Thanks, merged!

PR #27, opened 2016-07-28
Remove call to g_slice_set_config()

benesch:
I'm not clear on the divide between osxfuse/sshfs and libfuse/sshfs, but this
patch is necessary to fix a glib warning on OS X.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -3949,9 +3949,6 @@ int main(int argc, char *argv[])
 	if (!realpath(*exec_path, sshfs_program_path)) {
 		memset(sshfs_program_path, 0, PATH_MAX);
 	}
-
-	/* Until this gets fixed somewhere else. */
-	g_slice_set_config(G_SLICE_CONFIG_ALWAYS_MALLOC, TRUE);
 #endif /* __APPLE__ */
 	g_thread_init(NULL);
 


Nikratio:
Thanks for the patch! Could you include which glib warning this fixes, and why you think it's appropriate to just remove the call?

benesch:
Ah, sorry! The commit message has a far more detailed description, along with a link to the issue at osxfuse/sshfs: https://github.com/osxfuse/sshfs/issues/17

In short, the warning is:

'''
(process:5226): GLib-CRITICAL **: g_slice_set_config: assertion 'sys_page_size == 0' failed
SSHFS version 2.5 (OSXFUSE SSHFS 2.5.0)
'''

No one seems to know why this line of code was included in the first place—apparently it was in the original patch for OS X support—and removing it seems to have no negative effects. (I can confirm this myself; my SSHFS mount with this patch applied has been stable for over 24hrs.)

Nikratio:
Hm. Alright. If you add that info to the commit message, I'll be happy to merge.

Nikratio:
Oh yes, regarding libfuse/sshfs vs osxfuse/sshfs: the latter is supposed to be dormant and has been merged into libfuse/sshfs a few months ago,

benesch:
Sounds good! I think the commit message already contains most of that information—what in particular did you want me to add?

benesch:
Thank you much!

PR #53, opened 2017-04-03
Update the missing SSH options (that make sense) from ssh_config

### LINKED ISSUE #52: Update ssh_opts[] from options in `man ssh_config`
I note that the `CanonicalDomains` and `KexAlgorithms` options are missing from [`ssh_opts[]`](https://github.com/libfuse/sshfs/blob/master/sshfs.c#L296K).

I'm guessing that others may be missing, too.

Jakuje:
It is very confusing that some of the new ssh options are missing. I added the ones that are available in recent upstream OpenSSH release and make sense, including the KexAlgorithms, which is here for years.
Not sure about the usability of `CanonicalDomains` and other `Canonical*` options, which should take place in the configuration files. But adding them should not hurt (not included).
Resolves #52


Nikratio:
Thanks for the patch!  The problem here is really that a more recent sshfs may be used with an older sshfs. I think a better approach would be to avoid hardcoding that list altogether and simply pass everything to ssh - relying on it to report invalid options. What do you think?

Nikratio:
(Merging this request anyway, since it's clearly better than the status quo)

Jakuje:
Thanks. The OpenSSH will certainly report invalid options (in case it is new `sshfs` and old `openssh`), so it should not be a problem. The idea was to specify somewhat complete list that make sense (without removing deprecated options).
Allowing everything might result in some potential problems with future options or forwardings -- we have already passed [`ClearAllForwardings=yes` and `-x` and `-a`](https://github.com/libfuse/sshfs/blob/master/sshfs.c#L4007), which reflect to `ForwardX11=no` and `ForwardAgent=yes`. Specifying them once more on the command-line does not allow them, but other forwarding can still be added (well ... not sure why it was even disabled, but that is the same what [SFTP client does](https://github.com/openssh/openssh-portable/blob/master/sftp.c#L2292)).
We could on the other way create a list of "forbidden" options, if it would make sense to reject.

PR #55, opened 2017-05-15
Forward file locking requests to the sftp server

kalvdans:
This patch intercepts flock(2) system calls and forwards them to the remote host. The SFTP protocol extension is documented at https://github.com/openssh/openssh-portable/pull/66 .
Do not merge this pull request until the protocol is finalized by being merged into openssh.


Nikratio:
Thanks for working on this! This looks like a great feature to have, please keep me posted on progress on the protocol.

kalvdans:
Since the protocol extensions was rejected by the sftp-server team (see openssh/openssh-portable#66), we will have to close this pull request as well.

kalvdans:
We found an issue with this change; connection to servers without the flock extensions results in ENOSYS being returned for every flock() system call. It would be preferable to use the old behavior that uses kernel locking implementation.

PR #69, opened 2017-07-14
Removed nonempty option from man page

harrim4n:
The nonempty mount option hasn't been suported since libfuse < 3.0.0.
As sshfs links against libfuse 3.1.0, this man page entry is obsolete.


harrim4n:
Actually, I just noticed that some other options are not working as well, but are still in the man page.
Examples are cache_dir_timeout and cache_stat_timeout.
I'll try to remove all of them and will open a new pull request.

PR #70, opened 2017-07-14
Removed obsolete options from man page and FAQ

harrim4n:
I removed all obsolete options I could find (see #69).
I did this by simply trying out all options mentioned in the man page (except for the module ones).
All options that gave an error such as
```
fuse: unknown option(s): `-o OPTION'
```
were removed.

I think there might be a few options that were added since then which are not in the man page. I tried to create a list of them, but it wasn't clear to me which options were actually usable (and what they do), so I left it out. However,
```
sshfs --help
```
does print a list of supported options, which appears to be created by combing a static string and some function calls to libfuse functions (which in turn print some static strings).
Maybe these should be integrated into the man page?

### DIFF
--- a/FAQ
+++ b/FAQ
@@ -6,30 +6,26 @@ SSHFS FAQ
 2.     Advantage of sshfs over NFS and Samba
 3.     Create the device node
 4.     mv fails with "Operation not permitted"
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks, much appreciated!

PR #102, opened 2017-10-12
Fix memory leak on opendir error

tsavola:
buf_free(handle) is not necessary/possible before the free(handle) as buf_init is called at the end of sftp_request_wait, just before it succeeds.


Nikratio:
Thanks!

PR #103, opened 2017-10-15
Parallel connections

tsavola:
The -o max_conns=N option causes multiple SSH processes and response
processing threads to be created.  Each file handle is locked to the
connection which was used for opening it.  One-off requests can use any
connection.

The connection is chosen by checking the per-connection statistics:

  1. Choose connection with least outstanding requests; if it's a tie,
  2. choose connection with least directory handles; if it's a tie,
  3. choose connection with least file handles; if it's a tie,
  4. choose connection which has already been established.

The implementation assumes that the max_conns values will be small; it
uses linear search.

The primary purpose of this feature is to improve the responsiveness of a
file system during large file transfers.  It also helps in scenarios
where the bandwidth of a single connection is limited, but multiple
connections are allowed, and the aggregate bandwidth limit is larger.

https://sourceforge.net/p/fuse/mailman/message/36076333/

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -1756,8 +1756,9 @@ static struct conn *choose_conn()
 {
 	int best_index = 0;
 	uint64_t best_score = ~0ULL; /* smaller is better */
+	int i;
 
-	for (int i = 0; i < sshfs.max_conns; i++) {
+	for (i = 0; i < sshfs.max_conns; i++) {
 		uint64_t score = ((uint64_t) sshfs.conns[i].req_count << 43) +
 				 ((uint64_t) sshfs.conns[i].dir_count << 22) +
 				 ((uint64_t) sshfs.conns[i].file_count << 1) +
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! I'll look into it as soon as I can find some time.

Nikratio:
Sorry for the super-late response! Turns out this causes test errors when the feature is activated. Could you take a look? See current master branch.

Nikratio:
I have reverted master for now. Could you give me write permission for the pull request (https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)? Then I'll push a commit that activates a unit test for the feature.

tsavola:
I guess you don't need that anymore for this PR...

Nikratio:
No, but if you could still take a look at the test failure? Would very much like to merge this :-)

PR #118, opened 2018-03-20
New workaround renamexdev to enable moving files across remote FS

### LINKED ISSUE #100: mv fails with EPERM when moving a file between different remote FS's
Moving files on a same sshfs mount but between different FS's on the
remote host fails.  The server is OpenSSH 7.4p1-10+deb9u1 from Debian
with default config (sftp-server enabled).  The sshfs command is also
used with the default options; note that the option workaround=rename
does not change this error case:
```
$ sshfs -V
SSHFS version 2.8
FUSE library version: 2.9.7
fusermount version: 2.9.7
using FUSE kernel interface version 7.19
$ sshfs -o debug,sshfs_debug,loglevel=debug -f host:/ /home/user/sshfs/ | vipe
unique: 518, opcode: RENAME (12), nodeid: 27, insize: 114, pid: 20235
rename /mnt/fs1/dir1/file /mnt/fs2/dir2/file
[00074] EXTENDED
  [00074]         STATUS       28bytes (0ms)
   unique: 518, error: -1 (Operation not permitted), outsize: 16
$ strace mv file ../../fs2/dir2/ 2>&1 | egrep '^(rename|open|ioctl|read|write)\([^0-2]' | vipe
rename("file", "../../fs2/dir2/file") = -1 EPERM (Operation not permitted)
```

Relevant output of the sftp-server DEBUG3 log in /var/log/auth.log:
~~~~
Oct  4 11:47:03 localhost sftp-server[27991]: debug3: request 74: posix-rename
Oct  4 11:47:03 localhost sftp-server[27991]: posix-rename old "/mnt/fs1/dir1/file" new "/mnt/fs2/dir2/file"
Oct  4 11:47:03 localhost sftp-server[27991]: debug3: request 74: sent status 4
Oct  4 11:47:03 localhost sftp-server[27991]: sent status Failure
~~~~

Moving on the remote works as expected:
```
$ strace mv file ../../fs2/dir2/ 2>&1 | egrep '^(rename|open|ioctl|read|write)\([^0-2]' | vipe
rename("file", "../../fs2/dir2/file") = -1 EXDEV (Invalid cross-device link)
open("file", O_RDONLY|O_LARGEFILE|O_NOFOLLOW) = 3
open("../../fs2/dir2/file", O_WRONLY|O_CREAT|O_EXCL|O_LARGEFILE, 0600) = 4
ioctl(4, BTRFS_IOC_CLONE or FICLONE, 3) = -1 EXDEV (Invalid cross-device link)
read(3, "\nThe programs included with the "..., 131072) = 286
write(4, "\nThe programs included with the "..., 286) = 286
read(3, "", 131072)                     = 0
```

Here the error code returned by rename(3) is EXDEV which makes mv
attempt to actually move the data.  I wonder why POSIX does not allow
mv(1posix) to attempt a move after other kinds of failure.  What other
similar command is able to force a move of data in this case (while
preserving times and attributes) that is simpler to type than:
`bash -c 'cp -a "$@" && rm "${@:1:$#-1}"' arg0`

I suppose that sshfs returns EPERM instead of EXDEV because sftp does
not return a significant error code.  Would this be by design of sftp or
a shortcoming of OpenSSH's sftp-server?

As a fix maybe sshfs could always return EXDEV when rename fails and it
does not know precisely which error was triggered.  Or does sftp by
chance make it possible to detect mount points?

Regards

g-raud:
This is a fix that closes #100 (in the style of workaround as nothing better can be done it seems).

The change simply catches EPERM in sshfs_rename() and substitutes EXDEV.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -214,6 +214,7 @@ struct sshfs {
 	struct fuse_args ssh_args;
 	char *workarounds;
 	int rename_workaround;
+	int renamexdev_workaround;
 	int truncate_workaround;
 	int buflimit_workaround;
 	int unrel_append;
@@ -446,6 +447,8 @@ static struct fuse_opt workaround_opts[] = {
 	SSHFS_OPT("none",       fstat_workaround, 0),
 	SSHFS_OPT("rename",     rename_workaround, 1),
 	SSHFS_OPT("norename",   rename_workaround, 0),
+	SSHFS_OPT("renamexdev",   renamexdev_workaround, 1),
+	SSHFS_OPT("norenamexdev", renamexdev_workaround, 0),
 	SSHFS_OPT("truncate",   truncate_workaround, 1),
 	SSHFS_OPT("notruncate", truncate_workaround, 0),
 	SSHFS_OPT("buflimit",   buflimit_workaround, 1),
@@ -2333,6 +2336,8 @@ static int sshfs_rename(const char *from, const char *to, unsigned int flags)
 			}
 		}
 	}
+	if (err == -EPERM && sshfs.renamexdev_workaround)
+		err = -EXDEV;
 	return err;
 }
 
@@ -3365,6 +3370,7 @@ static void usage(const char *progname)
 "    -o workaround=LIST     colon separated list of workarounds\n"
 "             none             no workarounds enabled\n"
 "             [no]rename       fix renaming to existing file (default: off)\n"
+"             [no]renamexdev   fix moving across filesystems (default: off)\n"
 "             [no]truncate     fix truncate for old servers (default: off)\n"
 "             [no]buflimit     fix buffer fillup bug in server (default: on)\n"
 "             [no]fstat        always use stat() instead of fstat() (default: off)\n"
@@ -3851,6 +3857,7 @@ int main(int argc, char *argv[])
 #else
 	sshfs.rename_workaround = 0;
 #endif
+	sshfs.renamexdev_workaround = 0;
 	sshfs.truncate_workaround = 0;
 	sshfs.buflimit_workaround = 1;
 	sshfs.ssh_ver = 2;
--- a/sshfs.rst
+++ b/sshfs.rst
@@ -107,6 +107,8 @@ Options
 
    :rename: Emulate overwriting an existing file by deleting and
         renaming.
+   :renamexdev: Make rename fail with EXDEV instead of the default EPERM
+        to allow moving files across remote filesystems.
    :truncate: Work around servers that don't support truncate by
         coping the whole file, truncating it locally, and sending it
         back.
@@ -226,6 +228,17 @@ workaround=rename`. However, in this case it is still possible that
 someone (or something) recreates the destination file after SSHFS has
 removed it, but before SSHFS had the time to rename the old file. In
 this case, the rename will still fail.
+
+
+Permission denied when moving files across remote filesystems
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Most SFTP servers return only a generic "failure" when failing to rename
+across filesystem boundaries (EXDEV).  sshfs normally converts this generic
+failure to a permission denied error (EPERM).  If the option ``-o
+workaround=renamexdev`` is given, generic failures will be considered EXDEV
+errors which will make programs like `mv(1)` attempt to actually move the
+file after the failed rename.
 
 
 SSHFS hangs


sshfs.c
```diff
@@ -440,18 +441,20 @@ static struct fuse_opt sshfs_opts[] = {
...
+	SSHFS_OPT("norename",     rename_workaround, 0),
+	SSHFS_OPT("renamexdev",   renamexdev_workaround, 1),
+	SSHFS_OPT("norenamexdev", renamexdev_workaround, 0),
+	SSHFS_OPT("truncate",     truncate_workaround, 1),
+	SSHFS_OPT("notruncate",   truncate_workaround, 0),
+	SSHFS_OPT("buflimit",     buflimit_workaround, 1),
+	SSHFS_OPT("nobuflimit",   buflimit_workaround, 0),
+	SSHFS_OPT("fstat",        fstat_workaround, 1),
+	SSHFS_OPT("nofstat",      fstat_workaround, 0),
 	FUSE_OPT_END
```
Nikratio: Please remove all these spurious changes.

sshfs.rst
```diff
@@ -228,6 +230,16 @@ removed it, but before SSHFS had the time to rename the old file. In
 this case, the rename will still fail.
 
 
+Permission denied when moving files across remote filesystems
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SSH servers return a generic error when failing to rename across filesystem
```
Nikratio: Please add "some" (unless this is actually standard behavior, in that case please provide a link to the standard)

sshfs.rst
```diff
@@ -228,6 +230,16 @@ removed it, but before SSHFS had the time to rename the old file. In
...
+boundaries (EXDEV).  sshfs normally considers it a permission denied error
+(EPERM).  If the option `-o workaround=renamexdev` is given, generic failures
+will be considered EXDEV errors which will make a move program attempt to
```
Nikratio: Please use "programs like mv" instead of "a move program"

sshfs.rst
g-raud: This comes from the code (none of the 8 defined SSH_FX_* return codes describe EXDEV) and from the web at https://winscp.net/eng/docs/sftp_codes "Most SSH/SFTP servers support only SFTP version 3 that defines only codes 0 to 8." (Note that even amongst the "extended" error codes none describe EXDEV).  So I changed it to "Most SSH servers".

g-raud:
You may consider enabling the workaround by default, similar to what commit 491c1f2ba ("Map SSH2_FX_FAILURE to ENOTEMPTY for rmdir") did whereas `mkdir(3)` can fail with many different error conditions.

bjornfor:
What does this workaround mean? That the real fix is patching the OpenSSH SFTP server? What are the downsides to the sshfs workaround? (I just hit this issue on my system and wonder what the best way forward is.)

Nikratio:
Well, it means that sshfs will translate EPERM to EXDEV, so if the real problem is truly  EPERM, you will get an incorrect error code. And yes, a better fix would be for the server to return EXDEV :-)

g-raud:
On Mon, Aug 13, 2018 at 04:52:02PM -0700, Bjørn Forsman wrote:
> What does this workaround mean? That the real fix is patching the OpenSSH SFTP
> server? What are the downsides to the sshfs workaround? (I just hit this issue
> on my system and wonder what the best way forward is.)

There is no SFTP standard that enables an SSH server to return the
equivalent of the EXDEV error code.  The common error returned by the
server is translated to EPERM normally, which might or not correspond to
the real error on the server side; translating it to EXDEV makes the
client side program react differently, but once again the error might or
not correspond to the real error on the server side.

To improve things there are several ways:
1. Come up with a new SFTP standard that in a very long time will be
   implemented in most SFTP servers.
2. Implement an SFTP extension in one server (e.g. in OpenSSH) that
   allows to get a more precise error.
3. SFTP may make it possible to detect if there is a mountpoint between
   the remote source and the remote destination; so in case of an error
   only translate it to EXDEV if the remotes are on different mounts;
   this last case would only lower the number of inadequate EXDEV
   errors when the option "renamexdev" is on (to replace them by EPERM
   errors that may also be inadequate).

bjornfor:
Thanks for the info.

I found this patch interesting: https://github.com/openssh/openssh-portable/commit/f7fa706e70cc3d005acca995d022d65227185dcd. Unless I'm reading it wrong, openssh can be built with (undocumented) -DEXDEV flag since 2008, doing "the right thing" (IMHO) all by itself.

I realise that I don't even know what, in terms of syscalls, ```mv(1)``` does when it is asked to cross a filesystem boudary. I just found it odd that sshfs didn't operate transparently, like being able to "mv" files across filesystems in the same way as within filesystems.

PR #124, opened 2018-04-13
fix bug on create file

jessfraz:
Do two calls to `open_common` one for the create and once for the open.

So I don't think this is a correct fix but just opening here for some direction.

The problem is when running `touch file` on master returns:
```
touch: file: No such file or directory
```

It seems like Create and Open can't be done in one call.

Signed-off-by: Jess Frazelle <acidburn@microsoft.com>


jessfraz:
So sorry my explanation was terrible. So I was trying to fix a bug here: https://github.com/gl-prototypes/cmd-localfs/commit/d3e65e65e308c74dcd83143e592db22b20b9f81a

The behavior we were seeing is when the flag `O_CREAT` was on a `open` something like running `touch $file` or `echo "hello" > file` then it would error like so (but still create an empty file, which lead me to believe it just had to be done in two-step, once for `O_CREAT` and another to write):

```console
/local # ls
Dockerfile  LICENSE     Makefile    README.md   cmd         cmd-client  data        sshfs
/local # touch non_existent_file
touch: non_existent_file: No such file or directory
/local # touch non_existent_file
/local # echo "hello" > another_non_existent_file
sh: can't create another_non_existent_file: nonexistent directory
/local # echo "hello" > another_non_existent_file
/local # cat another_non_existent_file 
hello
/local # strace touch blah_another_file
execve("/bin/touch", ["touch", "blah_another_file"], 0x7fff641a7e28 /* 7 vars */) = 0
arch_prctl(ARCH_SET_FS, 0x7f21e8e9bb88) = 0
set_tid_address(0x7f21e8e9bbc0)         = 16
mprotect(0x7f21e8e98000, 4096, PROT_READ) = 0
mprotect(0x55c2e834d000, 16384, PROT_READ) = 0
getuid()                                = 0
utimensat(AT_FDCWD, "blah_another_file", NULL, 0) = -1 ENOENT (No such file or directory)
open("blah_another_file", O_RDWR|O_CREAT, 0666) = -1 ENOENT (No such file or directory)
write(2, "touch: blah_another_file: No suc"..., 52touch: blah_another_file: No such file or directory
) = 52
exit_group(1)                           = ?
+++ exited with 1 +++
```

jessfraz:
Then with this fix it works with no error:

```console
/local # ls
Dockerfile  LICENSE     Makefile    README.md   cmd         cmd-client  data        sshfs
/local # touch non_existent_file
/local # echo "hello" > another_non_existent_file
/local # cat another_non_existent_file 
hello
```

sshfs.c
```diff
@@ -2578,6 +2578,12 @@ static int sshfs_open_common(const char *path, mode_t mode,
 		g_free(sf);
 	}
 	buf_free(&buf);
+
+	if (fi->flags & O_CREAT){
+		fi->flags = fi->flags &~ O_CREAT;
+		return sshfs_open_common(path, mode, fi);
+	}
```
jessfraz: This is mostly the logic here, sorry the whitespace is a vim plugin

jessfraz:
And I definitely don't know if this fix is correct but that is the behavior :)

PR #127, opened 2018-06-02
Correct workaround condition for not using handle in sshfs_getattr

rianhunter:
In libfuse<3, when `fstat_workaround` was true, that meant to always
use the `path` argument to resolve fgetattr instead of the supplied
handle.  Before this change, the logic was interpreting
`fstat_workaround` to not use the `path` argument when it was
true. This change reverts to the libfuse<3 behavior.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -4,6 +4,7 @@ Unreleased Changes
 * Make utimens(NULL) result in timestamp "now" -- no more touched files
   dated 1970-01-01
 * New `createmode` workaround.
+* Fix `fstat` workaround regression.
 
 Release 3.3.2 (2018-04-29)
 --------------------------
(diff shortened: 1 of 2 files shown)


Nikratio:
Could you explain why one of the two is the better choice (leaving aside how it was done it the past)? Also, could you include a patch to the documentation?

Nikratio:
Also: thanks for tracking this down and reporting it!

rianhunter:
It's not simply a matter of defaults, there is no way to get the original workaround functionality with the code as it is without my proposed change. If you have a buggy sftp server that doesn't support fstat, you currently can't force sshfs to always use path-based stat.

Nikratio:
Got it, thanks. Could you please also update ChangeLog.rst though?

rianhunter:
No problem 👌

rianhunter:
Hmm, I made a new release commit to the best of my ability. Let me know if I did the correct thing.

Nikratio:
Thanks, but there's no need to make a new release. Just add something along the lines of "The fstat workaround is now working correctly again" to the top of the ChangeLog.rst file. No need to bump the version.

rianhunter:
Okay I have updated my repo with the corresponding ChangeLog.rst entry.

PR #128, opened 2018-06-08
Add workaround to always pass a 0 mode when creating a file

joshtriplett:
This works around servers that produce an error for any non-zero mode.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,8 @@
+Unreleased Changes
+------------------
+
+* New `createmode` workaround.
+
 Release 3.3.2 (2018-04-29)
 --------------------------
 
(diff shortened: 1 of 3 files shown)


joshtriplett:
This works around a [bug in the Windows port of OpenSSH](https://github.com/PowerShell/Win32-OpenSSH/issues/852).

Nikratio:
Thanks! Could you also update ChangeLog.rst?

joshtriplett:
Done. (I'd skipped ChangeLog.rst because I assumed that was updated as part of the release process, to avoid generating conflicts on every merge.)

Nikratio:
Thanks! If there are enough merges that I have to worry about conflict resolution in the changelog, then I would be very happy :-).

Issue #112, opened 2018-01-07
utimensat(NULL) not handled correctly


Nelyah:
I am not quite sure if this is a bug or if this is intended. The short version is, files created through the `touch` command have a creation date from 1970.

This is a cross-post from [Unix - StackExchange](https://unix.stackexchange.com/questions/415072/new-files-created-through-sshfs-date-from-1970) as I wasn't able to get an answer there.
I wish to mount a remote Linux directory to my local laptop using sshfs. I do this by using autofs. In `/etc/autofs/`, I therefore have in the `auto.ssh` file (I used the [Arch wiki][1] to guide me for the options I should use):

    cluster -fstype=fuse,rw,allow_other,IdentityFile=/home/mia/.ssh/id_rsa :sshfs\#mia@host\:

In the `auto.master` file I added the line:

    /home/mia/remote        /etc/autofs/auto.ssh    uid=1000,gid=1000,--ghost,--timeout=60
The folder is successfully mounted and its date is accurate:

     drwxr-xr-x  4 root  root     0 Jan  5 19:51 remote

However, if I create a new file inside the remote folder, the file's date is epoch (01-01-1970).

    $ cd remote/cluster
    $ touch test
    $ cp test test2
    $ ls -l test*
    -rw-r--r-- 1 mia mia    0 Jan  1  1970 test
    -rw-r--r-- 1 mia mia    0 Jan  7 14:07 test2

I have searched the man pages of mount, mount.fuse, sshfs, autofs but to no avail. I haven't been either able to determine if this was a bug or a feature. 

Both computers are up-to-date Linux Manjaro distribution.

  [1]: https://wiki.archlinux.org/index.php/SSHFS#Secure_user_access

Blackclaws:
Can reproduce. Also file editing does not update timestamps for me. Copying works as described.

quite:
I *cannot* reproduce this with fuse 2.9.4-1ubuntu3.1, sshfs 2.5-1ubuntu1 on Ubuntu 16.04 Xenial. If I install fuse3 from bkryza/fuse3 PPA, and compile sshfs from current git master, I do get the 1970-01-01 timestamp when issuing `touch`. Same result on a fresh Arch Linux, with fuse3 3.2.1, sshfs 3.3.1.

[...]

PR #130, opened 2018-06-15
Teach utimens to stamp with current time (#112)

quite:
This is a first attempt. It sets current time for both atime and mtime, iff both passed `timeval.tv_sec` were `0`.

Is this sensible? Are we supposed to emulate `utimensat` as close as possible? Nah... Anyway, the filexfer proto use here only supports seconds precision...

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Make utimens(NULL) result in timestamp "now" -- no more touched files
+  dated 1970-01-01
 * New `createmode` workaround.
 
 Release 3.3.2 (2018-04-29)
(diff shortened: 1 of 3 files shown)


Nikratio:
Looks like the right approach to me. I think you need to handle atime and mtime separately though, i.e. set them individually to the current time if they are 0.

quite:
LIke so..?

Nikratio:
Yes (I would drop the outer "if" statement completely, but I guess that's a matter of taste).

Now, if you want to make it perfect, add an entry to `ChangeLog.rst` and a test case to `test/test_sshfs.py`. You can probably model it after the `tst_chown` function. If both test and patch are correct, then the test should fail before your patch, and pass afterwards :-).

quite:
I think we got it :)

Nikratio:
Thanks!

Issue #77, opened 2017-08-06
Abort on network timeout


Nikratio:
We should take a look at the following patch, it may be worth merging:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=720455

Soukyuu:
That would probably solve issue #3

alexis57:
I have the same problem. Was the patch implemented? Is it a closed issue?

Nikratio:
Nope, this is still open. Pull requests welcome :-).

aerusso:
miklos@szeredi.hu did actually return my email (to the fuse mailing list back in 2013)


> Thanks for the patch.  How about using existing ssh options for this?
> 
>  -oServerAliveInterval=30
> 
> will disconnect after 90sec of network outage.  It will also disconnect when
> there's no sshfs activity, but together with -oreconnect that should work fine
> most of the time.
> 
> The problem with setting a default timeout is that there's no one value that's
> suitable for all users.  And the safest is no timeout, even though it may not be
> the most user friendly (ssh also has the timeout disabled by default).
> 
> Thanks,
> Miklos

I've been using those options since then, and it resolved my issue. IMO `-oServerAliveInterval=30` is a better default than no timeout, but it's a little late to change that default now.

smheidrich:
If that does indeed solve it, it would be nice if it could be explained in the documentation somewhere, perhaps as an extra section on network timeouts.

Nikratio:
@smheidrich Could you give it a shot? Happy to review pull requests :-).

smheidrich:
@Nikratio Sure, I'll try to write one this weekend :)

MAccadia:
I ran into this today. As I understand, the patch  (a9a2d75ab3db7e7fb12368ecf00f5e7599d46921) won't be merged because setting a timeout is not a good choice for everyone. However, it could be useful to have a warning in logs when sshfs is taking a long time.

In my case, it took quite some time to identify that the hanging was caused by sshfs. We use sshfs to mount $HOME ; everything was frozen and no clues in the logs.

Nikratio:
Pull requests are welcome :-).

MAccadia:
Unfortunately, I'm not a dev. Let's hope someone will have some time for that. It's not critical anyway.

PR #145, opened 2018-09-24
Docs: Add section on interrupted connections

smheidrich:
As promised in the comments on issue #77, this PR extends the docs to include the advice Miklos gave @aerusso in that e-mail (again in the comments on #77). Cf. also issue #3. The "SSHFS hangs" section was renamed to something more specific to differentiate it from this new section.

There are some outstanding questions / TODOs on my end before this can be merged (in addition to any you guys might have), maybe someone here can help:

1) If I have something mounted via SSHFS, remove the network cable and then go about my day ignoring the still-mounted directory, it *will* disconnect eventually after a couple of hours, so it's not correct to write that I/O will block indefinitely when the connection is interrupted. What causes this automatic disconnect if there is no timeout whatsoever? In the new doc section, I [wrote](https://github.com/libfuse/sshfs/compare/master...smheidrich:docs_on_interrupted_connections?expand=1#diff-4824d286673c49b341337b2300b31bdcR264) "[...] will block until the connection is either restored or closed altogether (e.g. manually)", and this additional reason should be added after "e.g. manually" if anyone can tell me what it is. OS resetting connections somehow?

2) Is it intended that ``fusermount -zu /path/to/mountpoint`` doesn't actually disconnect SSHFS (and therefore doesn't unblock any processes that were accessing something under the mountpoint), while e.g. ``killall sshfs`` does? I wrote about this in the new section because it's my experience, but if that is in turn just a bug (or just happens on my system), I might as well remove it.

3) Is it OK that I stress that all of these workarounds can lead to data loss, or does it sounds unnecessarily worrying? After all, just removing the network cable and not doing anything will eventually lead to the same thing anyway, cf. point (1). So maybe this just shouldn't be mentioned, people know what they're getting into with any type of network FS.

Thanks in advance, sorry the questions are longer than the actuall patch :-)

### DIFF
--- a/sshfs.rst
+++ b/sshfs.rst
@@ -243,8 +243,8 @@ errors which will make programs like `mv(1)` attempt to actually move the
 file after the failed rename.
 
 
-SSHFS hangs
-~~~~~~~~~~~
+SSHFS hangs for no apparent reason
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In some cases, attempts to access the SSHFS mountpoint may freeze if
 no filesystem activity has occured for some time. This is typically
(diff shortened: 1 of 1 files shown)


smheidrich:
@Nikratio Thanks for answering! Then I guess this is fine on my end (something more specific on point (1) can still be added later), if nobody else has any comments.

smheidrich:
Thanks for merging! :+1:

PR #156, opened 2019-01-04
Remap GID with -o idmap=user

Ratfink:
The manpage says that `-o idmap=user` maps the UID and GID, but it apparently only mapped the UID.  The code was in place to map the GID as well, but it was hidden behind `#ifdef __APPLE__`.  This PR removes those `#ifdef`s, and in a couple of cases the code inside an `#else`, to make the option behave as documented.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -4,6 +4,7 @@ Release 3.5.1 (2018-12-22)
 * Documentation updates
 * Build system updates
 * Added "BindInterface" as valid "-o" option.
+* Fixed "-o idmap=user" to map both UID and GID on all OSs.
 
 Release 3.5.0 (2018-08-28)
 --------------------------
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks! Could you please add a corresponding entry to `ChangeLog.rst` as well?

ChangeLog.rst
```diff
@@ -4,6 +4,7 @@ Release 3.5.1 (2018-12-22)
 * Documentation updates
 * Build system updates
 * Added "BindInterface" as valid "-o" option.
+* Fixed "-o idmap=user" to map both UID and GID on all OSs.
```
Nikratio: 3.5.1 is already released :-). You need to add it on top, for the next version.

Ratfink: Oh gosh, yes, how silly of me. 😏

Issue #157, opened 2019-01-10
Incompatible function pointer cast


mssalvatore:
The sshfs 2.10 build on Ubuntu 18.10 produces a compiler warning regarding an invalid function pointer cast:

> ../sshfs.c:1385:44: warning: cast between incompatible function types from ‘int (*)(void *, struct request *)’ to ‘gboolean (*)(void *, void *, void *)’ {aka ‘int (*)(void *, void *, void *)’} [-Wcast-function-type]

The offending line  is 
`g_hash_table_foreach_remove(sshfs.reqtab, (GHRFunc) clean_req, NULL);`

The definition of clean_req() is :
`static int clean_req(void *key_, struct request *req)`

GHRFunc is defined [here](https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#GHRFunc) as follows:
`gboolean
(*GHRFunc) (gpointer key,
            gpointer value,
            gpointer user_data);`

Even though we saw the warning in sshfs 2.10, it looks like this still applies to master.

Nikratio:
Thanks for the report! Would you be able to provide a patch? I don't have an Ubuntu system to test or reproduce. It looks as if it's as easy as changing `(GHRFunc) clean_req` to `(GHRFunc) &clean_req`.

mssalvatore:
Changing `(GHRFunc) clean_req to (GHRFunc) &clean_req` still produces the warning.

I think the issue is that the number of parameters of clean_req() doesn't match the spec. GHRFunc is defined as expecting 3 gpointers (void*), whereas clean_req(void *key_, struct request *req) only accepts 2 parameters.

Nikratio:
Right, sorry. In that case we'll need to add the extra argument to the function definition, e.g.

```
static int clean_req(void *key_, struct request *req, gpointer user_data) {
    (void) user_data;
    // [...]
```

PR #158, opened 2019-02-14
Update clean_req() to match the definition of GHRFunc

mssalvatore:
The current definition of the clean_req() function produces a compiler
warning as it expects 2 parameters but GHRFunc expects 3. This commit
resolves issue #157.


Nikratio:
Thanks!

PR #173, opened 2019-06-15
Added direct_io option

sunwire:
This option disables the use of page cache in kernel.
  This is useful for example if the file size is not known before reading it. 
  For example if you mount /proc dir from a remote host without the direct_io
  option, the read always will return zero bytes instead of actual data.

[root@local ~]# sshfs root@remotehost:/proc /remote/
[root@local ~]# cat /remote/uptime 
[root@local ~]# fusermount -u /remote 
[root@local ~]# sshfs -o direct_io root@remotehost:/proc /remote/
[root@local ~]# cat /remote/uptime 
2280959.53 2221494.84
[root@local ~]# fusermount -u /remote

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,13 @@
+Unreleased Changes
+------------------
+
+* Added "-o direct_io" option.
+  This option disables the use of page cache in kernel.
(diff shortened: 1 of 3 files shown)


Nikratio:
Thanks!

PR #176, opened 2019-06-29
Update sshfs.c

ggetsov:
verbose messages from ssh (-v option)


ggetsov:
This commit actually rises a deadlock problem if used with `-o password_stdin`

PR #187, opened 2019-09-22
Remedy gcc 7.4.0 warnings

kalvdans:
Two changes that makes sshfs compile without warnings on gcc 7.4.0.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -1740,7 +1740,7 @@ static void *sshfs_init(struct fuse_conn_info *conn,
 		sshfs.sync_read = 1;
 
 	// These workarounds require the "path" argument.
-        cfg->nullpath_ok = ~(sshfs.truncate_workaround || sshfs.fstat_workaround);
+        cfg->nullpath_ok = !(sshfs.truncate_workaround || sshfs.fstat_workaround);
 
         // Lookup of . and .. is supported
         conn->capable |= FUSE_CAP_EXPORT_SUPPORT;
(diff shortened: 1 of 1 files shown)


sshfs.c
```diff
@@ -3485,7 +3485,7 @@ static int sshfs_opt_proc(void *data, const char *arg, int key,
                                  * mountpoint to later be completely
                                  * unprivileged with libfuse >= 3.3.0.
                                  */
-                                sshfs.mountpoint = arg;
+                                sshfs.mountpoint = strdup(arg);
```
Nikratio: I don't think this is the right solution, it introduces a memory link instead.
Can't we simply declare sshfs.mountpoint as const?

kalvdans: Good point! I don't know if `fuse_opt_parse` promises to keep the strings it passes to `sshfs_opt_proc` alive until program exit. I was unable to find anything about lifetime in the documentation so better be safe than sorry and copy the value. We could of course free the previous value. What do you prefer?

Nikratio: You're right. Let's copy it and make sure that it's freed in the right places.

kalvdans: Ok, thanks for confirming! I will update the pull request with some `free` ;)

kalvdans: The code that assigns `sshfs.mountpoint` is inside an `if (!sshfs.mountpoint)` block. Thus it is only assigned to only once, and no memory leak can occur, and no free'ing needs to be added.

PR #189, opened 2019-10-26
Add build badge

fulldecent:


Nikratio:
Could you explain in what situation this is useful?

fulldecent:
This allows people to know that passing the build is important and that the project is high quality because it has automated builds set up.

Nikratio:
Sorry, I do not see how having automated builds setup shows that the project is high quality. Rather, it looks like providing free advertising for Travis. While I appreciate the service they are providing, I do not think that the SSHFS readme is the right place to place advertisements for a commercial company.

fulldecent:
Yes, that makes sense. And I also wish there was a more open and neutral way to set up builds.

PR #194, opened 2019-11-11
fix some whitespace and indentation

jeg139:


michaelforney:
Maybe a duplicate of #192. Perhaps we could combine our efforts?

sshfs.c
```diff
@@ -568,8 +568,8 @@ static inline void buf_init(struct buffer *buf, size_t size)
 		}
 	} else
 		buf->p = NULL;
-	buf->len = 0;
-	buf->size = size;
+		buf->len = 0;
+		buf->size = size;
```
michaelforney: I think this is wrong, there are no curly braces on the else statement, these two lines run in both cases.

sshfs.c
```diff
@@ -46,9 +46,9 @@
 #include <grp.h>
 #include <limits.h>
 #ifdef __APPLE__
-#  include <strings.h>
-#  include <libgen.h>
-#  include <darwin_compat.h>
+#include <strings.h>
+#include <libgen.h>
+#include <darwin_compat.h>
```
michaelforney: I don't think there is any issue here, the space is used as an indent to show that these files are included only when `__APPLE__` is defined.

jeg139:
Yes we can i was thinking #192 have already be pulled in. No problem.

sshfs.c
jeg139: Your alright i'll modify it.

sshfs.c
jeg139: Ok i'll modify it.

Nikratio:
I've just merged the older request. Could you please rebase on this one (if further changes are needed), or close this request?

Thanks for submitting this!

Nikratio:
Hmm. Looks like it still applied cleanly, nevermind!

PR #210, opened 2020-05-04
Disable symlinks client side when "follow_symlinks" enabled

SpiritsUnite:
When the "follow_symlinks" option is enabled, sshfs asks for the result
of stat instead of lstat. However, an untrusted server can still return
the result of lstat, and this can expose symlinks in the mount. So,
disable readlink in this case to disallow the use of these symlinks.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -2154,6 +2154,9 @@ static int sshfs_readlink(const char *path, char *linkbuf, size_t size)
 	if (sshfs.server_version < 3)
 		return -EPERM;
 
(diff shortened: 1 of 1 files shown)


Nikratio:
Thanks for the patch! I don't think clients should rely on there being no symlinks in the mountpoint if `follow_symlinks` is set. For example, if the symlink is dangling, SSHFS will not resolve it either. So I am not sure that this patch is a good idea...

SpiritsUnite:
Ah that's understandable - thanks for looking at this request anyway!

Issue #211, opened 2020-05-09
Seeking in directory stream has no effect


rozhuk-im:
Usual usage - works ok:
```
# mkdir ~/mnt/172.16.0.254/tmp/111
mkdir: /home/rim/mnt/172.16.0.254/tmp/111: File exists
Exit 1
rim@rimwks# rm -rf ~/mnt/172.16.0.254/tmp/111
rim@rimwks# mkdir ~/mnt/172.16.0.254/tmp/111
rim@rimwks# rm -rf ~/mnt/172.16.0.254/tmp/111
rim@rimwks# ls ~/mnt/172.16.0.254/tmp/
total 2228
drwxrwxrwt  1 root  wheel   384B May  9 09:30:31 2020 ./
drwxr-xr-x  1 root  wheel   1.0K Nov  3 23:37:03 2018 ../
drwx------  1 root  wheel    64B May  9 08:54:36 2020 mc-root/
srwxrwxrwx  1 88    wheel     0B May  9 06:53:04 2020 mysql.sock=
drwxr-xr-x  1 www   www     256B May  9 06:53:02 2020 nginx/
-rw-r-----  1 242   wheel   284B May  9 06:53:00 2020 prosody.err
-rw-r-----  1 242   wheel   2.2M May  9 09:30:17 2020 prosody.log
drwxrwxrwt  1 root  wheel     0B May  9 09:52:52 2020 vi.recover/
```

Log:

<details>

```
# sshfs -o sshfs_debug root@172.16.0.254:/ /home/rim/mnt/172.16.0.254/
SSHFS version 3.7.0
executing <ssh> <-x> <-a> <-oClearAllForwardings=yes> <-2> <root@172.16.0.254> <-s> <sftp>
Server version: 3
Extension: posix-rename@openssh.com <1>
Extension: statvfs@openssh.com <2>
Extension: fstatvfs@openssh.com <2>
Extension: hardlink@openssh.com <1>
Extension: fsync@openssh.com <1>
Extension: lsetstat@openssh.com <1>
[00001] LSTAT
  [00001]          ATTRS       41bytes (0ms)
[00002] EXTENDED
  [00002] EXTENDED_REPLY       97bytes (0ms)
[00003] OPENDIR
  [00003]         HANDLE       17bytes (0ms)
[00004] READDIR
[00005] READDIR
  [00004]           NAME     2193bytes (1ms)
  [00005]         STATUS       32bytes (1ms)
[00006] READDIR
[...]

PR #212, opened 2020-05-09
Improve FreeBSD support:

rozhuk-im:
- include posix_openpt() usage patch
- add workaround for readdir() issue: #211
- fix few warnings


sshfs.c
```diff
@@ -2338,22 +2382,9 @@ static int sshfs_readdir(const char *path, void *dbuf, fuse_fill_dir_t filler,
 		err = sftp_readdir_async(handle->conn, &handle->buf, dbuf,
 					 offset, filler);
 
-	return err;
-}
-
-static int sshfs_releasedir(const char *path, struct fuse_file_info *fi)
```
Nikratio, line 2344: Could you please move this function to the original location, so that the diff is easier to read?

sshfs.c
```diff
@@ -4173,6 +4205,7 @@ int main(int argc, char *argv[])
 	sshfs.truncate_workaround = 0;
 	sshfs.buflimit_workaround = 0;
 	sshfs.createmode_workaround = 0;
+	sshfs.readdir_workaround = 1;
```
Nikratio, line 4208: I think a workaround should always be disabled by default.

sshfs.rst
```diff
@@ -120,6 +120,13 @@ Options
    :buflimit: Work around OpenSSH "buffer fillup" bug.
    :createmode: Work around broken servers that produce an error when passing a
                 non-zero mode to create, by always passing a mode of 0.
+   :readdir: Work around file manager used that keeps dir open while
+             user add/remove files/dirs, that produce an error - all dirs
+             become empty for a while or until remount.
+             This happen because handle cached after opendir() but readdir()
+             does not use offset.
+             Workaround converts readdir() into "getdir()": opendir() and
+             releasedir() not exported to fuse; offset set to 0.
```
Nikratio, line 129: Hmm. Have you thought about storing the last last returned readdir option, and (on the next readdir) check if it corresponds to the stored value? If it matches, readdir can work as before. If it doesn't, it should log a warning and reset the stream. That behavior would make sense in all cases, so it wouldn't even need its own option.

rozhuk-im: Nope, this will not work.
https://github.com/openssh/openssh-portable/blob/4fa9e048c2af26beb7dc2ee9479ff3323e92a7b5/sftp-server.c#L1076
https://www.freebsd.org/cgi/man.cgi?query=readdir&apropos=0&sektion=0&manpath=FreeBSD+12.1-RELEASE+and+Ports&arch=default&format=html

After ```readdir()``` finish you can only ```closedir()```, that is why I enable this workaround for all platforms.
No way to use ```readdir()``` it other way except ```getdir()```.

sshfs.c
rozhuk-im: See below

sshfs.rst
Nikratio: Sorry, I don't follow. I am talking about FUSE's readdir(), not the one in SFTP. Here is the pseudo code:

```
void readdir(..., off_t pos) {
     static int last_offset;
     if (pos != last_offset) {
        log_warning("can't seek in readdir over SFTP, resetting stream to beginning")
        closedir(); opendir();
   }
    // readdir as before
   last_offset = <offset from last returned entry>
}    
```

Does this make more sense?

rozhuk-im: No, look at code in openssh, you will see that after ```readdir()``` handle cant be used any more.
On sshd server side it read all and send all to client, and no way to reset position on dir handle.
```offset``` other than 0 is not allowed in sshfs - there is few asserts for this.

Nikratio: What do you mean with "after readir()"? SFTP certainly allows for more than one readdir call (cf. https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-8.2.2). If this doesn't work in the given OpenSSH code, then this sounds like a bug to me (however, withI don't see the problem in the code on a cursory look).

rozhuk-im: Yes, it allow, but only until ```SSH_FX_EOF``` returned, after that dir handle can not be used any more.

Nikratio:
Are you still interested in working on this? If not, I'll close the pull request in a few days.

rozhuk-im:
FreeBSD already take this into ports tree :)

How do you suggest to rework this?
Remove option and make this "workaround" persistent?

lastthyme (approved):

sshfs.rst
Nikratio: Sorry, I don't think we are communicating well. I am saying that in the case where you have NOT encountered `SSH_FX_EOF` AND the FUSE client requests the next directory entry, there should be no need to close and re-open the stream.

Nikratio:
I'm happy with most of the patch, but I don't understand the need for the readdir workaround. I'd much rather see it removed from the patch and discussed separately.

Nikratio:
Are you still interested in working on this? If not, I'll close this pull request for now.

rozhuk-im:
> pam_mount

Yes: https://www.freshports.org/sysutils/pam_mount

PR #232, opened 2020-10-30
Rename option to 'passive' and add some example in manual.

dancerj:
Every time I want to use this command I seem to be looking up how to
do it.

`Slave` seems to have connotations I would rather use passive. 

Keep the old option so that we don't  break existing clients though.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -3636,7 +3636,7 @@ static void usage(const char *progname)
 "    -o ssh_protocol=N      ssh protocol to use (default: 2)\n"
 "    -o sftp_server=SERV    path to sftp server or subsystem (default: sftp)\n"
 "    -o directport=PORT     directly connect to PORT bypassing ssh\n"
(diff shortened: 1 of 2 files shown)


sshfs.c
```diff
@@ -3635,7 +3636,7 @@ static void usage(const char *progname)
 "    -o ssh_protocol=N      ssh protocol to use (default: 2)\n"
 "    -o sftp_server=SERV    path to sftp server or subsystem (default: sftp)\n"
 "    -o directport=PORT     directly connect to PORT bypassing ssh\n"
-"    -o slave               communicate over stdin and stdout bypassing network\n"
+"    -o passive               communicate over stdin and stdout bypassing network\n"
```
Nikratio, line 3639: Please fix alignment

Nikratio:
Thanks for the patch! Could you please fix the minor alignment issue above?

dancerj:
oops seems like I didn't actually make a change. Now I think I actually did, please take a look.

PR #240, opened 2021-01-09
Fix silently ignoring STATUS response after READ

### LINKED ISSUE #224: Read error on remote is reported as EOF
When trying to sequentially read a file (i.e. copying it) and the remote host runs into a read error, sshfs / fuse instead reports end-of-file. The local side has no way of knowing that an error occured and will assume EOF was actually reached. This results in partial copies without any error message whatsoever. This is especially bad in conjunction with `mv` (across filesystem boundaries), as that will unlink the source file after apparently successfully copying it.

`scp` properly reports the I/O error. Both local and remote host are running OpenSSH 8.3p1 on Arch, local has sshfs 3.7.0.

gralpli:
This attempts to fix #224, but is incomplete yet. Open questions:

- [ ] Does `res` need to be set in the immediately following else-if-branch and in the last else-branch as well? I guess not, because the case when `chunk->sio.error` equals `MY_EOF` is explicitly guarded a couple lines above.

- [ ] Do I need such a `MY_EOF` guard for the line that I added?

- [ ] When encountering an I/O error, SFTP reports a `SSH_FX_FAILURE` which, by the function `sftp_error_to_errno` is mapped to an `EPERM`. This leads to us getting an “Operation not permitted” over sshfs instead of an I/O error. Can `EPERM` be replaced by `EIO`?

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -3012,8 +3012,9 @@ static int wait_chunk(struct read_chunk *chunk, char *buf, size_t size)
 
 		if (rreq->res < 0) {
 			chunk->sio.error = rreq->res;
+			res = chunk->sio.error;
 			break;
-		} if (rreq->res == 0) {
+		} else if (rreq->res == 0) {
 			chunk->sio.error = MY_EOF;
 			break;
 		} else if (size < (size_t) rreq->res) {


Nikratio:
Thanks for the patch! I'm afraid I do not know the answers to your questions. The best I can do is to study the code carefully myself and try to figure this out, so this review might take a while. Any help is appreciated.

Once thing that comes to mind is that we probably need to check how real permissions errors are reported - we wouldn't want to get EIO if permissions are incorrect.

Nikratio:
Is this ready for review? I understand from the original message that there are still open questions, but it's not clear to me if you're still working on them (or other issues) or if you're waiting for someone to review this.

gralpli:
This is my first pull request where I’m actually fixing something and I don’t know how confident I need to be to release it. As far as I remember, this fixes the bug. But I can’t estimate all the implications that this code change has.

So I’d say yes, this is ready for review. The open questions are questions that a reviewer would need to answer to check if the code is correct.

Nikratio:
First step is to make sure the tests in the CI still pass, currently they are failing. That means either your code has a bug or the test has a bug - but in both cases it needs to be fixed.

PR #242, opened 2021-01-16
Workaround for mkdir on existing directory

peterbelm:
Added a secondary check so if a mkdir request fails with EPERM an access request will be tried - returning EEXIST if the access was successful. This matches the correct behaviour expected by applications such as git.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,11 @@
+Unreleased Changes
+--------------------------
+
+* Added a secondary check so if a mkdir request fails with EPERM an access request will be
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks! Could you please add an update to `ChangeLog.rst`?

peterbelm:
Should I be bumping the version or just leave the change at the top?

Nikratio:
Please add a new 'Unreleased Changes' section at the top, sorry for not mentioning that.

peterbelm:
No problem, all done.

breisig:
Can we get another release now this is merged?

PR #244, opened 2021-02-23
Fix deadlock in conn cleanup

thatguystone:
Calling through to request_free() from clean_req() causes deadlock since sshfs.lock is already held by the caller of clean_req().


Nikratio:
Thanks!

PR #245, opened 2021-02-23
Add option "idle_disconnect"

thatguystone:
I have a few sshfs mounts that I always have mounted, but only occasionally access. It feels a bit wasteful to maintain ssh connections so that they can just sit there.

This adds a new option, `-o idle_disconnect`, that causes a connection to disconnect once it has been idle for some number of seconds.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -341,6 +341,7 @@ struct sshfs {
 	int foreground;
 	int reconnect;
 	int delay_connect;
+	int idle_disconnect;
 	int passive;
 	char *host;
 	char *base_path;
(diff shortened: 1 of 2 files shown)


Nikratio:
Thanks for the patch! I am not sure I like the idea though. Have you confirmed what happens if the idle timeout is exceeded while an application has a file open? It seems to me that this would probably not play well with closing the connection.

In general, wouldn't this usecase be better served by mounting the sshfs filesystem on demand (with automount)?

thatguystone:
> Have you confirmed what happens if the idle timeout is exceeded while an application has a file open? It seems to me that this would probably not play well with closing the connection.

If any descriptor is still open, it falls through to `process_one_request()`, but I can always test it further

> In general, wouldn't this usecase be better served by mounting the sshfs filesystem on demand (with automount)?

That's a great point. I actually ended up here because automounters currently run sshfs as root, and that made me really uncomfortable. There's been some [discussion around fixing this](https://bugs.freedesktop.org/show_bug.cgi?id=73809), but so far, no movement. 

It would probably be a better option for me to get some movement going with systemd rather than working around it here.

Nikratio:
Ok, will close this now then. Thanks for reporting back!

thatguystone:
This was merged in https://github.com/systemd/systemd/pull/21315 and should be part of systemd 250.

PR #257, opened 2021-08-17
fix shellcheck identified issues, spelling and whitespace

a1346054:


Nikratio:
Thank you! The problem with white-space fixes is that they make git blame much harder to use and there is nothing that prevents trailing whitespace from being introduced again in the future. So I would prefer if you could revert those.

ccope:
Would you take the whitespace fixes if we added a CI step to check for them? I've been hacking on an unrelated feature myself and my editor keeps complaining about them.

Nikratio:
Yeah, I guess so. But please still put them in a separate commit.

a1346054:
I separated the pull request into individual commits, and additionally added an .editorconfig file so that trailing whitespace doesn't happen in the future.

ccope:
I'll look into setting up a github action today to enforce code style.

ccope:
I'd also hold off on merging this for the moment, we can bundle any/all formatting fixes together and then put the commit hashes in a .git-blame-ignore-revs file

Nikratio:
I don't think the `.editorconfig` file will be sufficient to ensure that this does not recur. There's too many editors that are not using it (including Emacs). 

Can we split this into one pull request for whitespace + ignore file + CI changes, and a second one for the other changes?

a1346054:
Split into #258

a1346054:
Rebased on latest master.

Leaving this PR open in case the .editorconfig file is considered useful.

Also still looking into how to add these particular fixes in an automated way.

kalvdans (approved):
I think this can be merged independently of a repo-wide reformatting and CI check.

Removed lines won't show up in git blame :)

PR #261, opened 2021-08-30
add .git-blame-ignore-revs

ccope:
Requires git >= 2.23. To enable this in your git client, run `git config --global blame.ignoreRevsFile .git-blame-ignore-revs`


ccope:
Note that you might still see ignored commits in git blame on lines that contained only whitespace.

PR #277, opened 2022-02-15
Implement connect to vsock.

g-easy:
"sshfs -o vsock=CID:PORT" will cause sshfs to connect directly to the
given vsock, bypassing ssh, and allowing high performance sshfs mounts
of a VM guest.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -50,6 +50,9 @@
 #  include <libgen.h>
 #  include <darwin_compat.h>
 #endif
+#ifdef __linux__
+#  include <linux/vm_sockets.h>
+#endif
 
 #include "cache.h"
 
@@ -349,6 +352,7 @@ struct sshfs {
 	pthread_mutex_t lock;
 	unsigned int randseed;
 	int max_conns;
+	char *vsock;
 	struct conn *conns;
 	int ptyfd;
 	int ptypassivefd;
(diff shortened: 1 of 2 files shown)


g-easy:
Example usage, doesn't even require a VM:

```
socat VSOCK-LISTEN:12345 EXEC:"/usr/lib/openssh/sftp-server",nofork
./sshfs -o vsock=2:12345 unused_host: ./tmp
```

(CID=2 means loopback)

Most of the performance gain is from reducing the amount of data copies (rather than avoiding ssh encryption), i.e. socat+nofork means the sftp-server's stdio is dup'd directly to the vsock.

g-easy:
> what sits on the other end of a vsock and implements the server-side protocol? Is it the kernel, or another userspace program?

For my use-case, it will be the usual userspace sftp-server, basically the same as in the socat+nofork example above.

> adding a capability to bypass SSH in a program called SSHFS does not seem wise

I understand where you're coming from. :) I have an existing setup where sshfs runs over ssh. I'd like to keep most of my setup and just make it run faster. i.e. I want to keep the sftp protocol part and change the plumbing.

I'd really like to be able to use a vsock directly from sshfs instead of having to plumb stdio around. sshfs already implements `-o directport` and `-o passive` but passive mode in particular disables multi-threading.

PR #301, opened 2024-04-22
Add back -o hard_remove flag

kalvdans:
See libfuse commit 8ee553dac029 ("fuse_new(): don't accept options that don't make sense for end-users") for details.

Our usecase depend on hard_remove behaviour, so in order to make it continue to work we have applied this patch. I hope it can be useful for others. Documentation is needed before merge.


kalvdans:
Tests are needed since the unit tests crashed with SIGSEGV when I forced hard_remove=1. Drafting.

PR #306, opened 2024-06-17
Fill stat info when returning cached data for readdir

jpalus:
Uncached and cached results for readdir were inconsistent -- the former returned correct stat info for directory entries while the latter didn't. That's because only names of entries were saved in cache without stat info. In turn this leads to issues like
https://github.com/junegunn/fzf/issues/3832 since directory traversal library (https://github.com/charlievieth/fastwalk in this case) relies on proper stat info returned by readdir. Hence when unchached result was returned it gave proper outcome, while with cached result it was wrong.

Cache stat info next to entry name to fix the issue. While file attributes are saved in cache already, they use full path as key. To avoid potentially plenty of allocations, string copying and cache lookups to get each attr, let's keep a copy of stat struct independently to be on the fast path.

### DIFF
--- a/cache.c
+++ b/cache.c
@@ -14,6 +14,7 @@
 #include <errno.h>
 #include <glib.h>
 #include <pthread.h>
+#include <sys/stat.h>
 
 #define DEFAULT_CACHE_TIMEOUT_SECS 20
 #define DEFAULT_MAX_CACHE_SIZE 10000
@@ -40,7 +41,7 @@ static struct cache cache;
 struct node {
 	struct stat stat;
 	time_t stat_valid;
-	char **dir;
+	GPtrArray *dir;
 	time_t dir_valid;
 	char *link;
 	time_t link_valid;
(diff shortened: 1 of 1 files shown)


h4sh5:
Thank you @jpalus!

I think having some test cases for cached stat would be great too, if you don't mind adding some into test/test_sshfs.py in another PR.

jpalus:
@h4sh5  I made an attempt at adding test today but Python's standard library isn't great for raw `readdir` results. There's [`os.scandir`](https://docs.python.org/3/library/os.html#os.scandir) which seems to fit perfectly except there's no way to detect missing stat info in `readdir` results: [`os.DirEntry`](https://docs.python.org/3/library/os.html#os.DirEntry) is already extremely limited in what it provides and the only two methods which could be used are worthless too:
- `stat()` always invokes system call so results from `readdir` are ignored anyway (see [docs](https://docs.python.org/3/library/os.html#os.DirEntry.stat))
- `is_file()` if it notices `d_type == DT_UNKNOWN` (which was the case for cached response prior to this PR) also results in separate `stat()` (see [source](https://github.com/python/cpython/blob/d24a22e9b6377797c3b602945347096fbe065670/Modules/posixmodule.c#L15719))

So while I've modified `tst_readdir` to use `scandir`, it doesn't really test anything new:
```python
    scandir_is = os.scandir(mnt_newdir)
    scandir_is = sorted(scandir_is, key=operator.attrgetter('name'))
    scandir_is2 = os.scandir(mnt_newdir)
    scandir_is2 = sorted(scandir_is2, key=operator.attrgetter('name'))
    scandir_is = list(map(lambda e: (e.name, e.is_file()), scandir_is))
    scandir_is2 = list(map(lambda e: (e.name, e.is_file()), scandir_is2))
    scandir_should = [(os.path.basename(file_), True), (os.path.basename(subdir), False)]
    scandir_should = sorted(scandir_should, key=operator.itemgetter(0))
    assert scandir_is == scandir_should
    assert scandir_is2 == scandir_should
```

Test passes on revision prior to this change even though it should not. Confirmed with `strace` in case of tests running with cache there are additional `stat()` calls.

h4sh5:
You could try to add tests that run the `stat` command or similar to get the details you need out of command output - just for safetly tho, use the absolate path to the commands in Ubuntu since the github actions tests run on that (e.g. /usr/bin/stat), and make sure you call subprocess with array arguments and `shell=False`

PR #342, opened 2026-01-31
Fix backslash escapes for `-o ssh_command`

jbellerb:
The option previously supported escaping spaces with backslash, but this was broken at some point. Spaces are correctly skipped, but the backslashes are never removed. As a result, it is impossible to use commands with arguments that include spaces.

Note that escaping requires a double backslash. The first backslash is always removed by `fuse_opt_parse`'s own escape handling.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -3931,6 +3931,7 @@ static char *tokenize_on_space(char *str)
 {
 	static char *pos = NULL;
 	char *start = NULL;
+	char *end = NULL;
 
 	if (str)
 		pos = str;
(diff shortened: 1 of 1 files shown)


abhinavagarwal07:
LGTM

Could you add a regression test in `test/test_sshfs.py`? Roughly: run sshfs with `-f -o sshfs_debug -o ssh_command='/nonexistent/wrap\\ per arg\\ with\\ space'` against a dummy `user@host:/` target, capture stderr, and assert the `executing …` line contains `<…/wrap per>` and `<arg with space>` as single tokens with no surviving backslashes. The bogus ssh_command makes exec fail fast, so it doesn't need passwordless localhost. Doubling the backslashes matters — `fuse_opt_parse` eats one layer before `tokenize_on_space` sees the value.

jbellerb:
Updated PR at #369. I accidentally made a bad force push after rebasing :(

PR #359, opened 2026-05-21
fix null-deref warning in tokenize_on_space, promote strict-warnings to required

abhinavagarwal07:
Remove redundant NULL guard in `tokenize_on_space()` that confused gcc's null-dereference analysis. `pos` is guaranteed non-NULL at that point (checked on line 3948, and `pos++` cannot produce NULL).

With the warning fixed, remove `continue-on-error: true` from the strict-warnings CI job so both gcc and clang strict-warning builds are now required.

Stacked on #353.

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -3954,7 +3954,7 @@ static char *tokenize_on_space(char *str)
 
 	start = pos;
 
-	while (pos && *pos != '\0') {
+	while (*pos != '\0') {
 		// break on space, but not on '\ '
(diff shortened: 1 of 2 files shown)


sshfs.c
abhinavagarwal07, line 3948: pos null check is done here already

abhinavagarwal07:
https://github.com/libfuse/sshfs/pull/353

h4sh5 (approved):

PR #371, opened 2026-06-29
fix: replace assert(offset==0) with early return in readdir

### LINKED ISSUE #338: On macOS, anything involving the Finder crashes sshfs
I had MacFUSE/sshfs working fine on my Apple M2 Macbook Pro running Sequoia 15.7.2. But, brew was confused, so I blew away my brew installation and reinstalled everything. I made sure that everything was ARM-based (it took a few tries).

I then reinstalled MacFUSE 5.1.2 and sshfs-3.7.5.pkg (both from [here](https://macfuse.github.io/)). And, my old scripts to start up sshfs mounting various Ubuntu boxes still work. I can use the command line to go into and interact with the mounted drives. I can even use BBEdit to edit files on the drive.

Problem: every time I try to use the Finder on an sshfs'd directory, sshfs crashes. I can view the drive's top level directory in the Finder, but anything deeper than that: boom. The same for using the standard File Open dialog from any app (e.g. when using "File/Open..." from BBEdit).

The entire Problem Reporter output is below, but the key seems to be that the thread that crashed is trying to use [`ssh_readdir()`](https://github.com/libfuse/sshfs/blob/master/sshfs.c#L2395):

```
Thread 6 Crashed:
0   libsystem_kernel.dylib        	       0x186e5a388 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x186e93848 pthread_kill + 296
2   libsystem_c.dylib             	       0x186d9c9e4 abort + 124
3   libsystem_c.dylib             	       0x186d9bc18 __assert_rtn + 284
4   sshfs                         	       0x10286e2c4 sshfs_readdir.cold.2 + 40
5   sshfs                         	       0x1028030f4 sshfs_readdir + 652
6   libfuse3.4.dylib              	       0x102f05090 0x102ef4000 + 69776
7   libfuse3.4.dylib              	       0x102f0e1c8 0x102ef4000 + 106952
8   libfuse3.4.dylib              	       0x102f0b014 fuse_session_process_buf_internal + 1272
9   libfuse3.4.dylib              	       0x102f07b88 0x102ef4000 + 80776
10  libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
11  libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8
```

Do you have any suggestions on how I could debug this?

P.S. Here's the complete Problem Reporter output:

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               sshfs [1783]
Path:                  /usr/local/bin/sshfs
Identifier:            sshfs
Version:               ???
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
Responsible:           iTerm2 [483]
User ID:               501

Date/Time:             2025-12-02 21:59:10.0389 -0500
OS Version:            macOS 15.7.2 (24G325)
Report Version:        12
Anonymous UUID:        C85561C1-847C-52E4-551C-190E6936CA25

Sleep/Wake UUID:       0EE9C91A-3E06-44CE-A9F8-560D6258D778

Time Awake Since Boot: 2500 seconds
Time Since Wake:       1211 seconds

System Integrity Protection: enabled

Crashed Thread:        6

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace SIGNAL, Code 6 Abort trap: 6
Terminating Process:   sshfs [1783]

Application Specific Information:
crashed on child side of fork pre-exec


Thread 0::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x186e51bb0 semaphore_wait_trap + 8
1   libdispatch.dylib             	       0x186cdd960 _dispatch_sema4_wait + 28
2   libdispatch.dylib             	       0x186cddf10 _dispatch_semaphore_wait_slow + 132
3   libfuse3.4.dylib              	       0x102f07530 fuse_session_loop_mt_312 + 284
4   libfuse3.4.dylib              	       0x102ef8fb0 fuse_loop_mt_312 + 84
5   libfuse3.4.dylib              	       0x102ef9108 fuse_loop_mt_31 + 56
6   sshfs                         	       0x1027ff608 main + 3056
7   dyld                          	       0x186af2b98 start + 6076

Thread 1:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 2:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 3:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   sshfs                         	       0x102805774 sftp_read + 76
2   sshfs                         	       0x10280543c process_requests + 168
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 4:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 5:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 6 Crashed:
0   libsystem_kernel.dylib        	       0x186e5a388 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x186e93848 pthread_kill + 296
2   libsystem_c.dylib             	       0x186d9c9e4 abort + 124
3   libsystem_c.dylib             	       0x186d9bc18 __assert_rtn + 284
4   sshfs                         	       0x10286e2c4 sshfs_readdir.cold.2 + 40
5   sshfs                         	       0x1028030f4 sshfs_readdir + 652
6   libfuse3.4.dylib              	       0x102f05090 0x102ef4000 + 69776
7   libfuse3.4.dylib              	       0x102f0e1c8 0x102ef4000 + 106952
8   libfuse3.4.dylib              	       0x102f0b014 fuse_session_process_buf_internal + 1272
9   libfuse3.4.dylib              	       0x102f07b88 0x102ef4000 + 80776
10  libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
11  libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 7:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 8:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 9:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 10:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8

Thread 11:
0   libsystem_kernel.dylib        	       0x186e527dc read + 8
1   libfuse3.4.dylib              	       0x102f0b6b8 0x102ef4000 + 95928
2   libfuse3.4.dylib              	       0x102f07ae4 0x102ef4000 + 80612
3   libsystem_pthread.dylib       	       0x186e93bc8 _pthread_start + 136
4   libsystem_pthread.dylib       	       0x186e8eb80 thread_start + 8


Thread 6 crashed with ARM Thread State (64-bit):
    x0: 0x0000000000000000   x1: 0x0000000000000000   x2: 0x0000000000000000   x3: 0x0000000000000000
    x4: 0x0000000000000000   x5: 0x000000000000002e   x6: 0x0000000000000002   x7: 0x0000000000000000
    x8: 0x1bcba9293acb183f   x9: 0x1bcba9285756283f  x10: 0x000000000000000a  x11: 0x0000000000000000
   x12: 0x0000000000000032  x13: 0x0000000000000000  x14: 0x0000000000000000  x15: 0x0000000000000010
   x16: 0x0000000000000148  x17: 0x00000001f5e7d548  x18: 0x0000000000000000  x19: 0x0000000000000006
   x20: 0x0000000000000d03  x21: 0x000000016d9d30e0  x22: 0x0000000000000016  x23: 0x00000001029a22de
   x24: 0x00000001f281f000  x25: 0x0000000000000003  x26: 0x0000000127605b80  x27: 0x0000000000000000
   x28: 0x0000000137606040   fp: 0x000000016d9d2a80   lr: 0x0000000186e93848
    sp: 0x000000016d9d2a60   pc: 0x0000000186e5a388 cpsr: 0x40001000
   far: 0x0000000000000000  esr: 0x56000080  Address size fault

Binary Images:
       0x1027fc000 -        0x1029affff sshfs (*) <bf8b19b7-7c53-3967-af89-da0fe38b0cd8> /usr/local/bin/sshfs
       0x102ef4000 -        0x102f1ffff libfuse3.4.dylib (*) <c28dc693-b389-3fcb-ab94-227c490a8c6d> /usr/local/lib/libfuse3.4.dylib
       0x186e51000 -        0x186e8c663 libsystem_kernel.dylib (*) <e5d90565-fa1a-3112-b048-59e321191677> /usr/lib/system/libsystem_kernel.dylib
       0x186cda000 -        0x186d2075f libdispatch.dylib (*) <63a54ce1-2ab6-3673-82e4-d8489d811475> /usr/lib/system/libdispatch.dylib
       0x186aec000 -        0x186b8757b dyld (*) <037bb3c1-5c6c-3ec9-af31-f5bded703b36> /usr/lib/dyld
               0x0 - 0xffffffffffffffff ??? (*) <00000000-0000-0000-0000-000000000000> ???
       0x186e8d000 -        0x186e99a77 libsystem_pthread.dylib (*) <022dc315-cf35-38da-939e-03800b5beff2> /usr/lib/system/libsystem_pthread.dylib
       0x186d24000 -        0x186da51f7 libsystem_c.dylib (*) <e098cb59-2c56-395c-ade1-6ef590e61199> /usr/lib/system/libsystem_c.dylib

External Modification Summary:
  Calls made by other processes targeting this process:
    task_for_pid: 0
    thread_create: 0
    thread_set_state: 0
  Calls made by this process:
    task_for_pid: 0
    thread_create: 0
    thread_set_state: 0
  Calls made by all processes on this machine:
    task_for_pid: 0
    thread_create: 0
    thread_set_state: 0

VM Region Summary:
ReadOnly portion of Libraries: Total=847.8M resident=0K(0%) swapped_out_or_unallocated=847.8M(100%)
Writable regions: Total=1.2G written=337K(0%) resident=337K(0%) swapped_out=0K(0%) unallocated=1.2G(100%)

                                VIRTUAL   REGION 
REGION TYPE                        SIZE    COUNT (non-coalesced) 
===========                     =======  ======= 
Kernel Alloc Once                   32K        1 
MALLOC                             1.2G       33 
MALLOC guard page                   96K        6 
STACK GUARD                       56.2M       12 
Stack                             14.4M       13 
Stack Guard                         16K        1 
VM_ALLOCATE                         16K        1 
__AUTH                             822K      161 
__AUTH_CONST                      12.8M      295 
__DATA                            11.2M      277 
__DATA_CONST                      10.8M      297 
__DATA_DIRTY                       453K       88 
__FONT_DATA                        2352        1 
__LINKEDIT                       617.1M        3 
__OBJC_RO                         61.4M        1 
__OBJC_RW                         2396K        1 
__TEXT                           230.7M      306 
__TPRO_CONST                       128K        2 
page table in kernel               337K        1 
shared memory                       32K        1 
===========                     =======  ======= 
TOTAL                              2.2G     1501 



-----------
Full Report
-----------

{"app_name":"sshfs","timestamp":"2025-12-02 21:59:10.00 -0500","app_version":"","slice_uuid":"bf8b19b7-7c53-3967-af89-da0fe38b0cd8","build_version":"","platform":1,"share_with_app_devs":0,"is_first_party":1,"bug_type":"309","os_version":"macOS 15.7.2 (24G325)","roots_installed":0,"incident_id":"34383CD0-56BC-4BEB-9A83-08C6CCF00AE8","name":"sshfs"}
{
  "uptime" : 2500,
  "procRole" : "Unspecified",
  "version" : 2,
  "userID" : 501,
  "deployVersion" : 210,
  "modelCode" : "Mac14,5",
  "coalitionID" : 533,
  "osVersion" : {
    "train" : "macOS 15.7.2",
    "build" : "24G325",
    "releaseType" : "User"
  },
  "captureTime" : "2025-12-02 21:59:10.0389 -0500",
  "codeSigningMonitor" : 1,
  "incident" : "34383CD0-56BC-4BEB-9A83-08C6CCF00AE8",
  "pid" : 1783,
  "translated" : false,
  "cpuType" : "ARM-64",
  "roots_installed" : 0,
  "bug_type" : "309",
  "procLaunch" : "2025-12-02 21:45:06.6156 -0500",
  "procStartAbsTime" : 40264837419,
  "procExitAbsTime" : 60507032613,
  "procName" : "sshfs",
  "procPath" : "\/usr\/local\/bin\/sshfs",
  "parentProc" : "launchd",
  "parentPid" : 1,
  "coalitionName" : "com.googlecode.iterm2",
  "crashReporterKey" : "C85561C1-847C-52E4-551C-190E6936CA25",
  "appleIntelligenceStatus" : {"state":"available"},
  "responsiblePid" : 483,
  "responsibleProc" : "iTerm2",
  "codeSigningID" : "sshfs",
  "codeSigningTeamID" : "3T5GSNBU6W",
  "codeSigningFlags" : 570503953,
  "codeSigningValidationCategory" : 6,
  "codeSigningTrustLevel" : 4294967295,
  "codeSigningAuxiliaryInfo" : 0,
  "instructionByteStream" : {"beforePC":"fyMD1f17v6n9AwCRm+D\/l78DAJH9e8Go\/w9f1sADX9YQKYDSARAA1A==","atPC":"AwEAVH8jA9X9e7+p\/QMAkZDg\/5e\/AwCR\/XvBqP8PX9bAA1\/WcAqA0g=="},
  "bootSessionUUID" : "3F0B4D44-5821-445E-92AF-60380200043A",
  "wakeTime" : 1211,
  "sleepWakeUUID" : "0EE9C91A-3E06-44CE-A9F8-560D6258D778",
  "sip" : "enabled",
  "exception" : {"codes":"0x0000000000000000, 0x0000000000000000","rawCodes":[0,0],"type":"EXC_CRASH","signal":"SIGABRT"},
  "termination" : {"flags":0,"code":6,"namespace":"SIGNAL","indicator":"Abort trap: 6","byProc":"sshfs","byPid":1783},
  "asi" : {"libsystem_c.dylib":["crashed on child side of fork pre-exec"]},
  "extMods" : {"caller":{"thread_create":0,"thread_set_state":0,"task_for_pid":0},"system":{"thread_create":0,"thread_set_state":0,"task_for_pid":0},"targeted":{"thread_create":0,"thread_set_state":0,"task_for_pid":0},"warnings":0},
  "faultingThread" : 6,
  "threads" : [{"id":37584,"threadState":{"x":[{"value":14},{"value":8589934595},{"value":171798697235},{"value":4410931413251},{"value":14680198217728},{"value":4410931412992},{"value":48},{"value":0},{"value":0},{"value":1},{"value":0},{"value":2},{"value":8403486872,"symbolLocation":0,"symbol":"vm_page_size"},{"value":105553175328945},{"value":4294967279},{"value":4344370322},{"value":18446744073709551580},{"value":8420582872},{"value":0},{"value":4955593392},{"value":4955593328},{"value":18446744073709551615},{"value":0},{"value":0},{"value":6129986576},{"value":105553163812864},{"value":4338703644,"symbolLocation":340,"symbol":"sshfs"},{"value":4338703304,"symbolLocation":0,"symbol":"sshfs"},{"value":4338700288,"symbolLocation":200,"symbol":"workaround_opts"}],"flavor":"ARM_THREAD_STATE64","lr":{"value":6556604768},"cpsr":{"value":1610616832},"fp":{"value":6129986432},"sp":{"value":6129986416},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558129072},"far":{"value":0}},"queue":"com.apple.main-thread","frames":[{"imageOffset":2992,"symbol":"semaphore_wait_trap","symbolLocation":8,"imageIndex":2},{"imageOffset":14688,"symbol":"_dispatch_sema4_wait","symbolLocation":28,"imageIndex":3},{"imageOffset":16144,"symbol":"_dispatch_semaphore_wait_slow","symbolLocation":132,"imageIndex":3},{"imageOffset":79152,"symbol":"fuse_session_loop_mt_312","symbolLocation":284,"imageIndex":1},{"imageOffset":20400,"symbol":"fuse_loop_mt_312","symbolLocation":84,"imageIndex":1},{"imageOffset":20744,"symbol":"fuse_loop_mt_31","symbolLocation":56,"imageIndex":1},{"imageOffset":13832,"symbol":"main","symbolLocation":3056,"imageIndex":0},{"imageOffset":27544,"symbol":"start","symbolLocation":6076,"imageIndex":4}]},{"id":37585,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4688200400},{"value":4},{"value":2},{"value":6131110740},{"value":0},{"value":5226172948},{"value":3},{"value":6131118246},{"value":0},{"value":0},{"value":8363565056,"symbolLocation":6840,"symbol":"read_memory"},{"value":4},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":4955593432},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6131117968},"sp":{"value":6131117888},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37586,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4688200400},{"value":4},{"value":2},{"value":6131684180},{"value":0},{"value":5226172948},{"value":3},{"value":6131691686},{"value":0},{"value":0},{"value":8363565056,"symbolLocation":6840,"symbol":"read_memory"},{"value":4},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":4955593528},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6131691408},"sp":{"value":6131691328},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37587,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":38772,"symbol":"sftp_read","symbolLocation":76,"imageIndex":0},{"imageOffset":37948,"symbol":"process_requests","symbolLocation":168,"imageIndex":0},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":5},{"value":4689249152},{"value":1103806595328},{"value":256},{"value":0},{"value":0},{"value":0},{"value":2563},{"value":2149818432},{"value":2149818432},{"value":41104},{"value":4339190304},{"value":0},{"value":16},{"value":3},{"value":8420587120},{"value":0},{"value":4689249136},{"value":105553145955072},{"value":6132264792},{"value":6132264791},{"value":5},{"value":4689249136},{"value":1209},{"value":4689249104},{"value":4338703824,"symbolLocation":520,"symbol":"sshfs"},{"value":4338703560,"symbolLocation":256,"symbol":"sshfs"}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4336932724},"cpsr":{"value":1610616832},"fp":{"value":6132264720},"sp":{"value":6132264656},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37590,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4687151696},{"value":4},{"value":2},{"value":6132831060},{"value":0},{"value":5226172948},{"value":3},{"value":6132838566},{"value":0},{"value":0},{"value":3},{"value":129},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":4690296984},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6132838288},"sp":{"value":6132838208},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37597,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4688200400},{"value":4},{"value":2},{"value":6133404500},{"value":0},{"value":5226172948},{"value":3},{"value":6133412006},{"value":0},{"value":0},{"value":8363565056,"symbolLocation":6840,"symbol":"read_memory"},{"value":4},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":4418715800},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6133411728},"sp":{"value":6133411648},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"triggered":true,"id":37601,"threadState":{"x":[{"value":0},{"value":0},{"value":0},{"value":0},{"value":0},{"value":46},{"value":2},{"value":0},{"value":2002880453841459263},{"value":2002880450025367615},{"value":10},{"value":0},{"value":50},{"value":0},{"value":0},{"value":16},{"value":328},{"value":8420578632},{"value":0},{"value":6},{"value":3331},{"value":6133985504},{"value":22},{"value":4338623198},{"value":8363569152,"symbolLocation":8,"symbol":"_OBJC_PROTOCOL_$_OS_dispatch_source_data_or"},{"value":3},{"value":4955593600},{"value":0},{"value":5224030272}],"flavor":"ARM_THREAD_STATE64","lr":{"value":6558398536},"cpsr":{"value":1073745920},"fp":{"value":6133983872},"sp":{"value":6133983840},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558163848,"matchesCrashFrame":1},"far":{"value":0}},"frames":[{"imageOffset":37768,"symbol":"__pthread_kill","symbolLocation":8,"imageIndex":2},{"imageOffset":26696,"symbol":"pthread_kill","symbolLocation":296,"imageIndex":6},{"imageOffset":494052,"symbol":"abort","symbolLocation":124,"imageIndex":7},{"imageOffset":490520,"symbol":"__assert_rtn","symbolLocation":284,"imageIndex":7},{"imageOffset":467652,"symbol":"sshfs_readdir.cold.2","symbolLocation":40,"imageIndex":0},{"imageOffset":28916,"symbol":"sshfs_readdir","symbolLocation":652,"imageIndex":0},{"imageOffset":69776,"imageIndex":1},{"imageOffset":106952,"imageIndex":1},{"imageOffset":94228,"symbol":"fuse_session_process_buf_internal","symbolLocation":1272,"imageIndex":1},{"imageOffset":80776,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}]},{"id":37602,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4688200400},{"value":4},{"value":2},{"value":0},{"value":0},{"value":5226172948},{"value":3},{"value":6134558886},{"value":0},{"value":0},{"value":8363565056,"symbolLocation":6840,"symbol":"read_memory"},{"value":4},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":5225071480},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6134558608},"sp":{"value":6134558528},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37603,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4689248768},{"value":4},{"value":2},{"value":0},{"value":0},{"value":5226172948},{"value":3},{"value":6135132326},{"value":0},{"value":0},{"value":0},{"value":2147483648},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":5224026520},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6135132048},"sp":{"value":6135131968},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37604,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4688200400},{"value":4},{"value":2},{"value":0},{"value":0},{"value":5226172948},{"value":3},{"value":6135705766},{"value":0},{"value":0},{"value":8363565056,"symbolLocation":6840,"symbol":"read_memory"},{"value":4},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":4418716056},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6135705488},"sp":{"value":6135705408},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37605,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4688200400},{"value":4},{"value":2},{"value":0},{"value":0},{"value":5226172948},{"value":3},{"value":6136279206},{"value":0},{"value":0},{"value":8363565056,"symbolLocation":6840,"symbol":"read_memory"},{"value":4},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":5224027384},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6136278928},"sp":{"value":6136278848},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}},{"id":37606,"frames":[{"imageOffset":6108,"symbol":"read","symbolLocation":8,"imageIndex":2},{"imageOffset":95928,"imageIndex":1},{"imageOffset":80612,"imageIndex":1},{"imageOffset":27592,"symbol":"_pthread_start","symbolLocation":136,"imageIndex":6},{"imageOffset":7040,"symbol":"thread_start","symbolLocation":8,"imageIndex":6}],"threadState":{"x":[{"value":4},{"value":0},{"value":4198400},{"value":1},{"value":4955594368},{"value":4},{"value":2},{"value":0},{"value":0},{"value":5226172948},{"value":3},{"value":6136852646},{"value":0},{"value":0},{"value":8363565056,"symbolLocation":6840,"symbol":"read_memory"},{"value":128},{"value":3},{"value":8420578608},{"value":0},{"value":5226172928},{"value":5225072344},{"value":1},{"value":4198400},{"value":0},{"value":4344463360},{"value":5226172948},{"value":5226173376},{"value":0},{"value":0}],"flavor":"ARM_THREAD_STATE64","lr":{"value":4344297144},"cpsr":{"value":1610616832},"fp":{"value":6136852368},"sp":{"value":6136852288},"esr":{"value":1442840704,"description":" Address size fault"},"pc":{"value":6558132188},"far":{"value":0}}}],
  "usedImages" : [
  {
    "source" : "P",
    "arch" : "arm64",
    "base" : 4336893952,
    "size" : 1785856,
    "uuid" : "bf8b19b7-7c53-3967-af89-da0fe38b0cd8",
    "path" : "\/usr\/local\/bin\/sshfs",
    "name" : "sshfs"
  },
  {
    "source" : "P",
    "arch" : "arm64",
    "base" : 4344201216,
    "size" : 180224,
    "uuid" : "c28dc693-b389-3fcb-ab94-227c490a8c6d",
    "path" : "\/usr\/local\/lib\/libfuse3.4.dylib",
    "name" : "libfuse3.4.dylib"
  },
  {
    "source" : "P",
    "arch" : "arm64e",
    "base" : 6558126080,
    "size" : 243300,
    "uuid" : "e5d90565-fa1a-3112-b048-59e321191677",
    "path" : "\/usr\/lib\/system\/libsystem_kernel.dylib",
    "name" : "libsystem_kernel.dylib"
  },
  {
    "source" : "P",
    "arch" : "arm64e",
    "base" : 6556590080,
    "size" : 288608,
    "uuid" : "63a54ce1-2ab6-3673-82e4-d8489d811475",
    "path" : "\/usr\/lib\/system\/libdispatch.dylib",
    "name" : "libdispatch.dylib"
  },
  {
    "source" : "P",
    "arch" : "arm64e",
    "base" : 6554566656,
    "size" : 636284,
    "uuid" : "037bb3c1-5c6c-3ec9-af31-f5bded703b36",
    "path" : "\/usr\/lib\/dyld",
    "name" : "dyld"
  },
  {
    "size" : 0,
    "source" : "A",
    "base" : 0,
    "uuid" : "00000000-0000-0000-0000-000000000000"
  },
  {
    "source" : "P",
    "arch" : "arm64e",
    "base" : 6558371840,
    "size" : 51832,
    "uuid" : "022dc315-cf35-38da-939e-03800b5beff2",
    "path" : "\/usr\/lib\/system\/libsystem_pthread.dylib",
    "name" : "libsystem_pthread.dylib"
  },
  {
    "source" : "P",
    "arch" : "arm64e",
    "base" : 6556893184,
    "size" : 528888,
    "uuid" : "e098cb59-2c56-395c-ade1-6ef590e61199",
    "path" : "\/usr\/lib\/system\/libsystem_c.dylib",
    "name" : "libsystem_c.dylib"
  }
],
  "sharedCache" : {
  "base" : 6553731072,
  "size" : 5039669248,
  "uuid" : "b395a5f6-c55f-3776-8d60-908ce43d7959"
},
  "vmSummary" : "ReadOnly portion of Libraries: Total=847.8M resident=0K(0%) swapped_out_or_unallocated=847.8M(100%)\nWritable regions: Total=1.2G written=337K(0%) resident=337K(0%) swapped_out=0K(0%) unallocated=1.2G(100%)\n\n                                VIRTUAL   REGION \nREGION TYPE                        SIZE    COUNT (non-coalesced) \n===========                     =======  ======= \nKernel Alloc Once                   32K        1 \nMALLOC                             1.2G       33 \nMALLOC guard page                   96K        6 \nSTACK GUARD                       56.2M       12 \nStack                             14.4M       13 \nStack Guard                         16K        1 \nVM_ALLOCATE                         16K        1 \n__AUTH                             822K      161 \n__AUTH_CONST                      12.8M      295 \n__DATA                            11.2M      277 \n__DATA_CONST                      10.8M      297 \n__DATA_DIRTY                       453K       88 \n__FONT_DATA                        2352        1 \n__LINKEDIT                       617.1M        3 \n__OBJC_RO                         61.4M        1 \n__OBJC_RW                         2396K        1 \n__TEXT                           230.7M      306 \n__TPRO_CONST                       128K        2 \npage table in kernel               337K        1 \nshared memory                       32K        1 \n===========                     =======  ======= \nTOTAL                              2.2G     1501 \n",
  "legacyInfo" : {
  "threadTriggered" : {

  }
},
  "logWritingSignature" : "473e8e6124cae7672dc266b90b64976919e61ccd",
  "trialInfo" : {
  "rollouts" : [
    {
      "rolloutId" : "645197bf528fbf3c3af54105",
      "factorPackIds" : {
        "SIRI_VALUE_INFERENCE_PERVASIVE_ENTITY_RESOLUTION" : "663e65b4a1526e1ca0e288a1"
      },
      "deploymentId" : 240000002
    },
    {
      "rolloutId" : "639124e81d92412bfb4880b3",
      "factorPackIds" : {

      },
      "deploymentId" : 240000012
    }
  ],
  "experiments" : [

  ]
}
}

Model: Mac14,5, BootROM 13822.41.1, proc 12:8:4 processors, 64 GB, SMC 
Graphics: Apple M2 Max, Apple M2 Max, Built-In
Display: Color LCD, 3024 x 1964 Retina, Main, MirrorOff, Online
Memory Module: LPDDR5, Hynix
AirPort: spairport_wireless_card_type_wifi (0x14E4, 0x4388), wl0: Jun 24 2025 05:20:23 version 23.40.31.0.41.51.179 FWID 01-b816d667
IO80211_driverkit-1485.12 "IO80211_driverkit-1485.12" Oct  7 2025 20:28:11
AirPort: 
Bluetooth: Version (null), 0 services, 0 devices, 0 incoming serial ports
Network Service: Wi-Fi, AirPort, en0
USB Device: USB31Bus
USB Device: USB31Bus
USB Device: USB31Bus
Thunderbolt Bus: MacBook Pro, Apple Inc.
Thunderbolt Bus: MacBook Pro, Apple Inc.
Thunderbolt Bus: MacBook Pro, Apple Inc.
```

Yike-Ye:
On macOS with macFUSE 5.3.x (FSKit backend) and macOS 27 Golden Gate beta, Finder/Spotlight may call readdir with a non-zero offset even when fuse_fill_dir is always called with offset=0 (old-style readdir). This triggers the assert and aborts the sshfs process.

Replace assert(offset == 0) with if (offset != 0) return 0 in:
- sftp_readdir_async() in sshfs.c
- sftp_readdir_sync() in sshfs.c
- cache_readdir() in cache.c

Since all entries are returned in the first (offset=0) call, returning 0 on subsequent calls is semantically correct and prevents the crash.

Fixes: https://github.com/libfuse/sshfs/issues/338
Tested: macOS 27 Golden Gate beta 2, macFUSE 5.3.2

### DIFF
--- a/cache.c
+++ b/cache.c
@@ -386,7 +386,13 @@ static int cache_readdir(const char *path, void *buf, fuse_fill_dir_t filler,
 	struct node *node;
 	struct cache_dirent **cdent;
 
-	assert(offset == 0);
+#ifndef __APPLE__
+	assert(offset == 0);  /* keep for non-macOS; see sshfs.c sftp_readdir_async */
+#endif	
+	/* On macOS with macFUSE 5.3.x / FSKit, may be called with offset != 0.
+	 * All entries were already returned in the first (offset=0) call. */
+	if (offset != 0)
+		return 0;
 
 	pthread_mutex_lock(&cache.lock);
 	node = cache_lookup(path);
--- a/sshfs.c
+++ b/sshfs.c
@@ -2318,7 +2318,21 @@ static int sftp_readdir_async(struct conn *conn, struct buffer *handle,
 
 	int done = 0;
 
-	assert(offset == 0);
+	/*
+	 * sshfs uses old-style readdir (filler always called with offset=0),
+	 * so FUSE should return all entries in one call (offset=0).
+	 * On macOS with macFUSE 5.3.x / FSKit backend, Finder or Spotlight
+	 * may call readdir again with a non-zero offset even in old-style mode.
+	 * When that happens, all entries were already returned in the first
+	 * call, so we can safely return 0 (no more entries) instead of crashing.
+	 * See: https://github.com/libfuse/sshfs/issues/338
+	 */
+#ifndef __APPLE__
+ assert(offset == 0);
+#endif
+	if (offset != 0)
+		return 0;
+
 	while (!done || outstanding) {
 		struct request *req;
 		struct buffer name;
@@ -2390,7 +2404,13 @@ static int sftp_readdir_sync(struct conn *conn, struct buffer *handle,
 			     void *buf, off_t offset, fuse_fill_dir_t filler)
 {
 	int err;
+	/* See comment in sftp_readdir_async for why offset != 0 is handled
+	 * this way instead of asserting. */
+#ifndef __APPLE__
 	assert(offset == 0);
+#endif	
+ if (offset != 0)
+		return 0;
 	do {
 		struct buffer name;
 		err = sftp_request(conn, SSH_FXP_READDIR, handle, SSH_FXP_NAME, &name);


cache.c
dtgriscom, line 1: Why assert that offset must be 0, and then handle the case where it isn't?

Yike-Ye: Because previously sshfs uses `buf_get_entries` ([sshfs.c: 985](https://github.com/libfuse/sshfs/blob/master/sshfs.c#L961-L993)) with offset=0, and `assert(offset=0)` is then used in `sftp_readdir_async` / `sftp_readdir_sync` for bug detection. In this process, all directory entries  are returned from the server within one step. If the offset is non-zero, something might be wrong. See the notes in [fuse.h](https://github.com/macfuse/library/blob/afdd74cfcea87e7f6b2f6dc202fb6d575caedf12/include/fuse.h#L669C1-L697C3): 

```
	/** Read directory
	 *
	 * The filesystem may choose between two modes of operation:
	 *
	 * 1) The readdir implementation ignores the offset parameter, and
	 * passes zero to the filler function's offset.  The filler
	 * function will not return '1' (unless an error happens), so the
	 * whole directory is read in a single readdir operation.
	 *
	 * 2) The readdir implementation keeps track of the offsets of the
	 * directory entries.  It uses the offset parameter and always
	 * passes non-zero offset to the filler function.  When the buffer
	 * is full (or an error happens) the filler function will return
	 * '1'.
	 *
	 * When FUSE_READDIR_PLUS is not set, only some parameters of the
	 * fill function (the fuse_fill_dir_t parameter) are actually used:
	 * The file type (which is part of stat::st_mode) is used. And if
	 * fuse_config::use_ino is set, the inode (stat::st_ino) is also
	 * used. The other fields are ignored when FUSE_READDIR_PLUS is not
	 * set.
	 */
	FUSE_DARWIN_EXTEND_OPERATION(
		readdir,
		int (*) (const char *, void *, fuse_fill_dir_t, off_t,
			 struct fuse_file_info *, enum fuse_readdir_flags),
		int (*) (const char *, void *, fuse_darwin_fill_dir_t, off_t,
			 struct fuse_file_info *, enum fuse_readdir_flags)
	)
```
However, the updated MacFuse uses the new FSKit backend instead of kernel extension backend. This will sometimes deliver a non-zero `offset`, and thus lead to abort and disconnection.

dtgriscom: But won't the assert cause the code to bail on a non-zero offset? Do you perhaps want to #ifdef the assert so it isn't active on macOS?

Yike-Ye: I have replaced the `assert(offset == 0)` with `if (offset !=0) return 0`, which will return early instead of aborting the sshfs process. Although this may affect the FSKit backend behavior I mentioned, it works well for me so far. By the way, the `#ifdef` flag suggestion is a good point. I will update the PR shortly.

Yike-Ye: Oh, I got what you meant in the beginning. There was a typo and have been fixed in the update =)

dmik:
Please see https://github.com/macfuse/macfuse/issues/1131#issuecomment-4884964594 for some (failing) test results.

Yike-Ye:
@dmik Thanks for testing and for the very clear repro — you were right, PR #371 only removed the assertion, it didn't fix the underlying problem.

**Root cause.** sshfs uses old-style `readdir` on top of a *single-pass* SFTP directory handle: it reads the handle once, streaming entries to FUSE. On macOS with macFUSE 5.3.x (the FSKit backend), the kernel re-issues `readdir` on the *same* handle — either with a non-zero offset, or an `offset==0` rewind. By then the SFTP handle is exhausted, so the re-read yields nothing. PR #371 turned that into `return 0`, i.e. an empty listing. And because sshfs has a directory cache (`cache.c`, default 20s), that empty result gets cached — so the whole directory looks empty for ~20s until the cache expires. That matches exactly what you saw: contents vanish under load and reappear once the pressure (and the cache) goes away.

**Fix.** Instead of streaming from the single-pass handle, snapshot the entire directory into memory on the first `readdir` and serve every subsequent call — any offset, including rewinds — from that snapshot. This makes `readdir` idempotent: no matter how often or with what offset the kernel re-issues it, it always gets the complete listing, so the cache never stores a partial/empty result.

**Testing.** I reproduced your scenario on a real remote mount (an HPC filesystem over a ProxyJump) and drove it hard with VS Code (large project + .git, global search, file watchers), with logging added to the cache layer. With the `return 0` build the directory goes empty; with the snapshot build every `readdir` returns the full listing, there are zero empty results, and the directory never disappears under sustained load.

Branch: https://github.com/Yike-Ye/sshfs/tree/fix/readdir-full-snapshot

@h4sh5 — happy to open a separate PR for this if you'd like to take it. Otherwise I'm fine leaving #371 as-is for now: macFUSE is actively reworking this area for macOS 27 support, so things may shift and I'd rather avoid conflicts until that settles. Just let me know which you prefer.

dmik:
@Yike-Ye thanks for the proper fix! I tested it with my stress test and it passes. The directory contents survives and everything else seems to function well. A few notes:

1. I test everything with the kernel backend, not FSKit (I don't use it due to various limitations). But I guess that it doesn't matter since the underlying logic changes that bring offset != 0 live in the macOS kernel and affect both backends.

2. I had to update to macFUSE 5.3.3 (released just yesterday) because of common daemonization (fork) issues after touching new ObjC machinery via `MFMount.framework` (i.e. using `libfuse` before the first fork I suppose). With 5.3.2, running `sshfs` without `-f` would eventually crash in the ObjC/fork guard. I think `sshfs` daemonization should be changed so that it happens before triggering `MFMount` use because macFUSE 5.3.3 now always forces `sshfs` to run in foreground (`fuse: daemonize requested after mount started; continuing in foreground`) to avoid this crash.

3. While you mention that vanished directories could be related to caching, it was not the case here. First, I use the cache timeout of 1 second (not the default 20) and second, I tried to disable caching completely while testing (didn't help with the VS Code test case).

Yike-Ye:
@dmik Thanks for testing and confirming it passes — great to have independent verification on the kernel backend. And you're right that the offset != 0 behaviour comes from the macOS kernel, so it affects both backends.

Thanks especially for the correction on caching — you're right, it's not the cause. The real cause is `readdir` itself returning empty: sshfs reads the directory over a single-pass SFTP handle, and when the kernel re-issues `readdir` on that already-exhausted handle (non-zero offset, or an offset==0 rewind), the re-read yields nothing. The cache sits *above* readdir and just re-serves its answer, so lowering the timeout or disabling it entirely doesn't help — the underlying readdir keeps returning empty. That matches your result. The snapshot fix addresses it at the source: the whole directory is captured in memory on the first readdir and every later call is served from that snapshot, so readdir is idempotent and never returns empty, regardless of cache settings.

On daemonization — agreed, and macFUSE 5.3.3 already moved that way on the libfuse side: it defers the actual mount until first session use and avoids forking in `fuse_daemonize()` after the mount has started, which is why sshfs is now forced to foreground. That's a libfuse-level change, independent of this fix.

Framing for anyone following: #371 and the snapshot branch are two versions of the *same* readdir fix, not separate issues.
- **#371 (`return 0`)** is a minimal stopgap — it stops the crash (the old `assert(offset==0)` aborting under the new kernel behaviour) but returns an empty listing, so directories still vanish under load.
- **The snapshot branch** is the proper fix — readdir becomes idempotent and always returns the full listing. It supersedes #371.

@h4sh5 two ways to land this, your call:
- **Preferred:** I replace this PR (or open a fresh one) with the snapshot fix.
- **Or:** merge #371 as the interim crash-stop, and I'll open the snapshot fix as a follow-up.

For clarity, this is entirely within sshfs's readdir callback (`sshfs.c`) and doesn't touch libfuse or macFUSE. macFUSE isn't at fault — 5.3.x legitimately started issuing readdir with non-zero offsets / rewinds (which FUSE has always permitted), and that simply exposed a long-standing assumption in sshfs that offset is always 0.

bfleischer:
When enumerating a directory, the FSKit API does not provide the number of directory entries it expects us to return (or a buffer size for us to fill). This means macFUSE does not know how many directory entries to request from libfuse (and by extension the file system server). This means the macFUSE FSKit backend may request more directory entries than FSKit can handle during a particular call. This results in the next readdir call re-requesting directory entries that have already been returned in the previous response. This might be related to the issue.

Yike-Ye:
@bfleischer Thanks, that's a really helpful explanation — it accounts for the non-zero `offset` we see once a request goes through the FSKit layer. As mentioned in my previous comment, I've addressed this by snapshotting the whole directory in memory and serving every `readdir` from it, so the listing stays correct no matter how often or at what offset it gets re-requested.

Since this stems from the FSKit API itself, I'd imagine it won't need major changes on the macFUSE side, so the snapshot commit should be able to serve as a long-term fix here. Either way, we'll see how the sshfs maintainer wants to proceed.

bfleischer:
@Yike-Ye I have not had time to look into this more closely yet because I have been busy with macFUSE 5.3, but it looks like either macFUSE or sshfs is breaking the FUSE contract here.

The fact that sshfs is affected could mean that other file systems are affected as well. If so, I may need to add a workaround in macFUSE. However, if sshfs is the one breaking the FUSE contract and the issue is isolated to sshfs, then a workaround on the macFUSE side may not be necessary.

dmik:
@Yike-Ye Some more details. It appears that something's still wrong with your latest fix. Under my stress test (VS Code running and open on a large .git project), the directory contents is reported by `sshfs` inconsistently. First, newly created files don't appear in the listing (although they do get created just fine on the remote). Second, the size of a file touched by VS Code alternates between consequent directory reloads in `far2l` from a real value to a smaller one. In VS Code this looks like the file gets suddenly (externally) truncated. Reloading it there restores the contents. The file is definitely not changed on the remote side. The moment I close VS Code this weird behavior disappears and newly created files appear as well.

When I get back to stock sshfs-3.7.6 (i.e. without your patches at all) directory contents expectedly disappears, but I don't see this weird file size behavior. I can freely edit files in VS Code (files themselves are readable and writable, it's only the directory listing that gets empty) and their sizes remain consistent.

PS. Forgot to mention that I somehow don't see offset == 0 assertions in stock sshfs-3.7.6 with macFUSE 5.3.3 at all. Maybe there are some related changes that affect that but still, it's strange. We are missing something.

PPS. Using sshfs 2.10 with macFUSE 5.3.3 shows nor crashes neither missing directory contents. And no weird file size behavior when editing files in VS Code. So the problem of sshfs 3.x may be bigger than just this offset issue… I recalled now that I tried sshfs 3.x some months ago (before I switched to macFUSE 5.2 and even before switching to macOS 26 at all) and I got annoyed by disappearing directories so I went back to sshfs 2.10 back then (until the recent round).

Yike-Ye:
@dmik Thanks for the follow-up — that repro pinned it down. Both symptoms trace back to my first patch: it snapshotted the directory once per `opendir` and then served every `readdir` from that frozen snapshot. While an app keeps the directory open (VS Code's watcher), the snapshot was never refreshed — so new files never showed up, and the stale stat sizes in it fought the live sizes from `getattr`, which is exactly the truncation / size flip-flop you saw. Once VS Code is closed the handle is dropped, the next `opendir` builds a fresh snapshot, and everything reappears.

I've fixed this in 30bf21b by making the snapshot **per-enumeration** instead of per-opendir. On `offset == 0` (a fresh listing or a rewinddir) it now reopens a fresh SFTP handle and re-reads the directory from the server, so new files and updated attributes show up; `offset != 0` is still served from the snapshot taken at offset 0, so the listing never comes back empty on an already-exhausted handle.

I tested it on localhost with a process holding a single directory handle open (to mimic the VS Code case): new files appear and file sizes update within the same open session, and concurrent listings stay complete and stable. Could you run it through your stress test?

Branch: https://github.com/Yike-Ye/sshfs/tree/fix/readdir-full-snapshot

Let me know if there is any issue on this fix.

Yike-Ye:
@dmik On macFUSE 5.3.3 I think the crash this PR guards against and the recent `fuse_daemonize()` change are **one root cause with two symptoms**.

macFUSE 5.3.3 now refuses to fork inside `fuse_daemonize()` once the mount has started (`daemonize requested after mount started; continuing in foreground`), as noted in the [5.3.3 release notes](https://macfuse.github.io/2026/07/04/macfuse-5.3.3.html). The reason is that the subsystems the mount brings up — XPC, DiskArbitration, CoreFoundation, and libdispatch — are **fork-unsafe** on macOS. Following sshfs's upstream order `mount → connect → daemonize`, `fuse_daemonize()`'s `fork()` carries that already-initialized state into the daemon child **in an undefined post-fork state**. The FSKit→FUSE enumeration bridge running on top of that dispatch/XPC machinery then loses its cookie/verifier bookkeeping and either forwards a bogus non-zero offset down to sshfs's old-style readdir (the crash this PR handles) or aborts outright. The 5.3.3 notes point right at this: the mount is now deferred until after daemonization specifically to keep "XPC, DiskArbitration, CoreFoundation, and MFMount state from being created before daemonization."

In other words, the problem was never *being in the background* — it's *forking a process that already holds the mount state*. As long as the fork happens before `fuse_mount()` brings those subsystems up, the offset stays 0 and nothing crashes.

To keep sshfs backgrounding as before, I reordered the macOS-only path (`#ifdef __APPLE__`) to `connect → daemonize → mount`:
- `ssh_connect()` runs first, in the foreground, so the interactive password/passphrase prompt still reaches the terminal;
- `fuse_daemonize()` forks next, while the process is still clean;
- `fuse_mount()` runs last, so XPC/CF/FSKit are initialized **fresh in the final daemon process** rather than inherited across the fork.

This still daemonizes — it just forks *before* the mount, so those subsystems are never crossed by a fork. Same principle as 5.3.3 deferring the mount until after daemonization. The non-Apple path is left untouched to avoid any Linux regression.

Confirmed working on macOS 27 + macFUSE 5.3.3: two mounts (one key-auth, one keyboard-interactive/password) both daemonize to the background as before, keep their SSH connections alive, and enumerate directories with no offset crash — stable across 2h+ of uptime.

Change: https://github.com/Yike-Ye/sshfs/commit/b85d744

bfleischer:
### Regarding the `assert(offset == 0)` or `if (offset != 0)`:

sshfs is breaking the FUSE contract here. When using readdir option 1, sshfs may be called with non-zero offsets, but sshfs may chose to ignore the offset and return the whole directory contents in one go by passing 0 as offset to the filler function.

```
	 * 1) The readdir implementation ignores the offset parameter, and
	 * passes zero to the filler function's offset.  The filler
	 * function will not return '1' (unless an error happens), so the
	 * whole directory is read in a single readdir operation.
```

`assert(offset == 0)` or simply returning 0 on `offset == 0` (without any directory entries) is a bug or at the very least a wrong assumption in sshfs. This is true for macOS and Linux. On Linux the issue might never be triggered, though.

### Regarding macFUSE 5.3.3:

macFUSE 5.3 uses an XPC connection as channel to exchange FUSE messages with the FSKit backend. Unlike file descriptors, XPC connections do not survive a `fork()`.

> The FSKit→FUSE enumeration bridge running on top of that dispatch/XPC machinery then loses its cookie/verifier bookkeeping and either forwards a bogus non-zero offset down to sshfs's old-style readdir (the crash this PR handles) or aborts outright. 

This is not correct. The XPC connection is severed by forking the process. That's the main reason libfuse in macFUSE 5.3.3 delays mounting the volume (and by extension establishing the XPC connection). FSKit is not affected by sshfs forking and there is no bogus data being sent to sshfs. 

The FUSE contract states that calling readdir with `offset != 0` is valid. The two readdir modes are local to libfuse, which means the macOS and Linux kernels, as well as FSKit, do not know which mode a particular file system chooses.

libfuse sits between the kernel, or FSKit, and the file system process, and it handles both modes internally. From the kernel’s and FSKit’s perspective, readdir looks the same either way.

It is valid for libfuse to pass readdir calls with offset != 0 to sshfs, and sshfs must handle them correctly. If passing `offset != 0` to sshfs wasn't legal, this would rather be a libfuse bug than a kernel or FSKit bug.

@Yike-Ye 

> To keep sshfs backgrounding as before, I reordered the macOS-only path (#ifdef __APPLE__) to connect → daemonize → mount:

This should not be necessary. `fuse_main()` internally calls `fuse_daemonize()` after `fuse_mount()`. `fuse_daemonize()` after `fuse_mount()` is the de facto standard. This way the mount parameters are checked and errors can be printed before daemonizing the process.

Not having to reorder `fuse_daemonize()` and `fuse_mount()` in every file system is the whole point of delayed mounting in macFUSE 5.3.3.

### Where does this leave us?

This PR and more precisely "replace assert(offset==0) with early return in readdir" does not fix the real issue. That's why I'm closing it. 

Snapshotting or simply re-reading the whole directory might be better options. Let's continue the discussion in https://github.com/libfuse/sshfs/issues/338.

Issue #373, opened 2026-07-06
Call `fuse_daemonize` before `fuse_mount`


dmik:
The current order affects macFUSE 5.2.0+ which switched to ObjC `MFMount.framework` in `libfuse`. Using `fork` after touching some ObjC objects and then touching them again in the child after `fork` is not supported on macOS and triggers its own `fork` guard that terminates the `sshfs` process:

```
Thread 3 Crashed:
0   libsystem_kernel.dylib        	       0x180efeb10 __abort_with_payload + 8
1   libsystem_kernel.dylib        	       0x180f260e8 abort_with_payload_wrapper_internal + 104
2   libsystem_kernel.dylib        	       0x180f26080 abort_with_reason + 32
3   libobjc.A.dylib               	       0x180b105ec _objc_fatalv(unsigned long long, unsigned long long, char const*, char*) + 128
4   libobjc.A.dylib               	       0x180b1056c _objc_fatal(char const*, ...) + 44
5   libobjc.A.dylib               	       0x180af4b10 performForkChildInitialize(objc_class*, objc_class*) + 396
6   libobjc.A.dylib               	       0x180ada7d0 initializeNonMetaClass + 620
7   libobjc.A.dylib               	       0x180af6258 initializeAndMaybeRelock(objc_class*, objc_object*, locker_mixin<lockdebug::lock_mixin<objc_lock_base_t>>&, bool) + 176
8   libobjc.A.dylib               	       0x180ad9f8c lookUpImpOrForward + 288
9   libobjc.A.dylib               	       0x180ad9b84 _objc_msgSend_uncached + 68
10  MFMount                       	       0x1035d2f0c specialized static Bridge.log(level:function:_:) + 564
11  MFMount                       	       0x1035dfa9c closure #1 in MFChannelCopyNextMessage(_:) + 232
12  MFMount                       	       0x1035df984 MFChannelCopyNextMessage + 32
13  libfuse3.4.dylib              	       0x102718880 0x102700000 + 100480
14  libfuse3.4.dylib              	       0x1027140b4 0x102700000 + 82100
15  libsystem_pthread.dylib       	       0x180f35c58 _pthread_start + 136
16  libsystem_pthread.dylib       	       0x180f30c1c thread_start + 8

…

Thread 3 crashed with ARM Thread State (64-bit):
    x0: 0x0000000000000008   x1: 0x0000000000000001   x2: 0x0000000000000000   x3: 0x0000000000000000
[...]

PR #374, opened 2026-07-06
macOS: do not call fuse_session_fd before fuse_daemonise

dmik:
This appears to be the right sequence and also fixes ObjC/fork issues on macOS with macFUSE 5.2.0+ (see #373).

### DIFF
--- a/sshfs.c
+++ b/sshfs.c
@@ -4518,7 +4518,7 @@ int main(int argc, char *argv[])
 		exit(1);
 	}
 
-#if !defined(__CYGWIN__)
+#if !defined(__CYGWIN__) && !defined(__APPLE__)
 	res = fcntl(fuse_session_fd(se), F_SETFD, FD_CLOEXEC);
 	if (res == -1)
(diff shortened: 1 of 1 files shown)


bfleischer:
Please see my comment https://github.com/libfuse/sshfs/issues/373#issuecomment-4889678146

sshfs.c
bfleischer, line 4522: This line forces macFUSE 5.3.3 and later to materialize the delayed mount. Starting with macFUSE 5.3.3, fuse_daemonize() will no longer perform the fork(), after the mount has been materialized.

dmik:
Please also back port it to the 2.x branch. I tested the patch locally and it works well with macFUSE 5.3.3.

It's also worth removing this from `sshfs.c` in the 2.x branch:

```c
#ifdef __APPLE__
#  include <fuse_darwin.h>
#endif
```

as macFUSE 5.x doesn't have it (and doesn't need it).

h4sh5:
replaced with #380 as that seems more appropriate

